PR #43, opened 2021-03-17
add azure_iot.h

danewalton:
Noteable changes from the `sdk_spike` branch:
- Changed the result from `AzureIoTError_t` to `AzureIoTResult_t`.
- Naming updates for some parameters to fit the FreeRTOS guidelines.
- Added a brief description for the file doxygen header.
- Updated enum values from uppercase snake case to Pascal prefixed with `e`.
    - ex: `AZURE_IOT_SUCCESS ` to `eAzureIoTSuccess`

### DIFF
--- a/source/include/azure_iot.h
+++ b/source/include/azure_iot.h
@@ -128,9 +128,10 @@ AzureIoTResult_t AzureIoT_MessagePropertiesFind( AzureIoTMessageProperties_t * p
  * @param[in] ulMessageSize The length of \p pucMessage.
  * @param[in] pucBuffer An intermediary buffer to put the base64 decoded key.
  * @param[in] ulBufferLength The length of \p pucBuffer.
- * @param[out] ppucOutput The buffer into which the resulting HMAC256 hashed, base64 encoded message will
+ * @param[out] pucOutput The buffer into which the resulting HMAC256 hashed, base64 encoded message will
  * be placed.
- * @param[out] pulOutputLength The output length of \p ppucOutput.
+ * @param[in] ulOutputSize Size of \p pucOutput.
+ * @param[out] pulOutputLength The output length of \p pucOutput.
  * @return An #AzureIoTResult_t with the result of the operation.
  */
 AzureIoTResult_t AzureIoT_Base64HMACCalculate( AzureIoTGetHMACFunc_t xAzureIoTHMACFunction,
@@ -140,7 +141,8 @@ AzureIoTResult_t AzureIoT_Base64HMACCalculate( AzureIoTGetHMACFunc_t xAzureIoTHM
                                                uint32_t ulMessageSize,
                                                uint8_t * pucBuffer,
                                                uint32_t ulBufferLength,
-                                               uint8_t ** ppucOutput,
+                                               uint8_t * pucOutput,
+                                               uint32_t ulOutputSize,
                                                uint32_t * pulOutputLength );
 
 #endif /* AZURE_IOT_H */
--- a/source/include/azure_iot.h
+++ b/source/include/azure_iot.h
@@ -34,11 +34,11 @@
 
 typedef enum AzureIoTResult
 {
-    eAzureIoTSuccess = 0,        /**< Success. */
-    eAzureIoTInvalidArgument,    /**< Input argument does not comply with the expected range of values. */
-    eAzureIoTStatusOutOfMemory,  /**< The system is out of memory. */
-    eAzureIoTStatusItemNotFound, /**< The item was not found. */
-    eAzureIoTFailed,             /**< There was a failure. */
+    eAzureIoTSuccess = 0,     /**< Success. */
+    eAzureIoTInvalidArgument, /**< Input argument does not comply with the expected range of values. */
+    eAzureIoTOutOfMemory,     /**< The system is out of memory. */
+    eAzureIoTItemNotFound,    /**< The item was not found. */
+    eAzureIoTFailed,          /**< There was a failure. */
 } AzureIoTResult_t;
 
 /**
@@ -61,16 +61,16 @@ typedef uint32_t ( * AzureIoTGetHMACFunc_t )( const uint8_t * pucKey,
                                               uint32_t ulDataLength,
                                               uint8_t * pucOutput,
                                               uint32_t ulOutputLength,
-                                              uint32_t * pucBytesCopied );
+                                              uint32_t * pulBytesCopied );
 
 /**
- * @brief Initialize Azure IoT
+ * @brief Initialize Azure IoT middleware.
  *
  */
 AzureIoTResult_t AzureIoT_Init();
 
 /**
- * @brief Deinitialize Azure IoT
+ * @brief Deinitialize Azure IoT middleware.
  *
  */
 void AzureIoT_Deinit();
@@ -90,7 +90,7 @@ AzureIoTResult_t AzureIoT_MessagePropertiesInit( AzureIoTMessageProperties_t * p
                                                  uint32_t ulBufferLength );
 
 /**
- * @brief Append a property name and value.
+ * @brief Append a property name and value to a message.
  *
  * @param[in] pxMessageProperties The #AzureIoTMessageProperties_t* to use for the operation.
  * @param[in] pucName The name of the property to append.
@@ -111,7 +111,7 @@ AzureIoTResult_t AzureIoT_MessagePropertiesAppend( AzureIoTMessageProperties_t *
  * @param[in] pxMessageProperties The #AzureIoTMessageProperties_t* to use for the operation.
  * @param[in] pucName The name of the property to find.
  * @param[in] nameLength Length of the property name.
- * @param[out] ppucOutValue The output pointer to the value.
+ * @param[out] ppucOutValue The output pointer to the property value.
  * @param[out] pulOutValueLength The length of \p ppucOutValue.
  * @return An #AzureIoTResult_t with the result of the operation.
  */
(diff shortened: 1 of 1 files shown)


source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ * AZURE_IOT_DO_NOT_USE_CUSTOM_CONFIG macro should not be defined. */
+#ifndef AZURE_IOT_DO_NOT_USE_CUSTOM_CONFIG
+    /* Include custom config file before other headers. */
+    #include "azure_iot_config.h"
+#endif
+
+/* Include config defaults header to get default values of configs not
+ * defined in azure_iot_mqtt_config.h file. */
+#include "azure_iot_config_defaults.h"
+
```
ericwol-msft, line 30: why are there blank lines between the includes?

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+#include "FreeRTOS.h"
+
+#include "azure/az_iot.h"
+
+typedef enum AzureIoTResult
+{
+    eAzureIoTSuccess = 0,        /**< Success. */
+    eAzureIoTInvalidArgument,    /**< Input argument does not comply with the expected range of values. */
+    eAzureIoTStatusOutOfMemory,  /**< The system is out of memory. */
```
ericwol-msft, line 39:
```suggestion
    eAzureIoTOutOfMemory,  /**< The system is out of memory. */
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+    eAzureIoTStatusItemNotFound, /**< The item was not found. */
```
ericwol-msft, line 40:
```suggestion
    eAzureIoTItemNotFound, /**< The item was not found. */
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+
+typedef uint64_t ( * AzureIoTGetCurrentTimeFunc_t )( void );
+
+typedef uint32_t ( * AzureIoTGetHMACFunc_t )( const uint8_t * pucKey,
+                                              uint32_t ulKeyLength,
+                                              const uint8_t * pucData,
+                                              uint32_t ulDataLength,
+                                              uint8_t * pucOutput,
+                                              uint32_t ulOutputLength,
+                                              uint32_t * pucBytesCopied );
```
ericwol-msft, line 64:
```suggestion
                                              uint32_t * pulBytesCopied );
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * @brief Initialize Azure IoT
```
ericwol-msft, line 67:
```suggestion
 * @brief Initialize Azure IoT middleware
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ *
+ */
+AzureIoTResult_t AzureIoT_Init();
+
+/**
+ * @brief Deinitialize Azure IoT
```
ericwol-msft, line 73:
```suggestion
 * @brief Deinitialize Azure IoT middleware
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @param[out] pxMessageProperties The #AzureIoTMessageProperties_t* to use for the operation.
+ * @param[out] pucBuffer The pointer to the buffer.
+ * @param[in] ulWrittenLength The length of the properties already written (if applicable).
+ * @param[in] ulBufferLength The length of \p pucBuffer.
+ * @return An #AzureIoTResult_t with the result of the operation.
+ */
+AzureIoTResult_t AzureIoT_MessagePropertiesInit( AzureIoTMessageProperties_t * pxMessageProperties,
+                                                 uint8_t * pucBuffer,
+                                                 uint32_t ulWrittenLength,
```
ericwol-msft, line 89: swap ulWrittenLength and ulBufferLength? keeps buf len next to the buffer

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+                                                 uint32_t ulBufferLength );
+
+/**
+ * @brief Append a property name and value.
```
ericwol-msft, line 93:
```suggestion
 * @brief Append a property name and value to a message.
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+                                                   const uint8_t * pucValue,
+                                                   uint32_t ulValueLength );
+
+/**
+ * @brief Find a property in the message property bag.
+ *
+ * @param[in] pxMessageProperties The #AzureIoTMessageProperties_t* to use for the operation.
+ * @param[in] pucName The name of the property to find.
+ * @param[in] nameLength Length of the property name.
+ * @param[out] ppucOutValue The output pointer to the value.
```
ericwol-msft, line 114:
```suggestion
 * @param[out] ppucOutValue The output pointer to the property value.
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+AzureIoTResult_t AzureIoT_MessagePropertiesFind( AzureIoTMessageProperties_t * pxMessageProperties,
+                                                 const uint8_t * pucName,
+                                                 uint32_t ulNameLength,
+                                                 const uint8_t ** ppucOutValue,
+                                                 uint32_t * pulOutValueLength );
+
+/**
+ * @brief As part of symmetric key authentication, HMAC256 a buffer of bytes and base64 encode the result.
+ *
+ * @note This is used within Azure IoT Hub and Device Provisioning APIs should a symmetric key be set.
```
ericwol-msft, line 127:
```suggestion
 * @note This is used within Azure IoT Hub and Device Provisioning APIs using a symmetric key.
```

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @param[in] xAzureIoTHMACFunction The #AzureIoTGetHMACFunc_t function to use for HMAC256 hashing.
+ * @param[in] pucKey A pointer to the base64 encoded key.
+ * @param[in] ulKeySize The length of the \p pucKey.
+ * @param[in] pucMessage A pointer to the message to be hashed.
```
ericwol-msft, line 132:
```suggestion
 * @param[in] pucMessage A pointer to the blob to be hashed.
```
?

ericwol-msft (approved):

source/include/azure_iot.h
danewalton: separate logical includes from different places

source/include/azure_iot.h
danewalton: I think the logic is it goes smallest to greatest. Sort of defining boundaries from 0 to max

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @param[out] pxMessageProperties The #AzureIoTMessageProperties_t* to use for the operation.
+ * @param[out] pucBuffer The pointer to the buffer.
+ * @param[in] ulWrittenLength The length of the properties already written (if applicable).
+ * @param[in] ulBufferLength The length of \p pucBuffer.
+ * @return An #AzureIoTResult_t with the result of the operation.
+ */
+AzureIoTResult_t AzureIoT_MessagePropertiesInit( AzureIoTMessageProperties_t * pxMessageProperties,
+                                                 uint8_t * pucBuffer,
+                                                 uint32_t ulWrittenLength,
```
jspaith, line 89: Nit - consider (only) ulAlreadyWrittenLength or ulOffset (don't like that as much).  Normally when I see a "written" I assume that the function called will set the # of bytes that it has written and this won't be a marker for something some other entity (even if it was a previous call to same function) has written.

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+                                                 uint32_t ulBufferLength );
+
+/**
+ * @brief Append a property name and value to a message.
+ *
+ * @param[in] pxMessageProperties The #AzureIoTMessageProperties_t* to use for the operation.
+ * @param[in] pucName The name of the property to append.
```
jspaith, line 96: What rules, if any, apply to URL Encoding of these property name/value?  We should call out either if it's required or else optional.

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ * @param[in] ulMessageSize The length of \p pucMessage.
+ * @param[in] pucBuffer An intermediary buffer to put the base64 decoded key.
+ * @param[in] ulBufferLength The length of \p pucBuffer.
+ * @param[out] pucOutput The buffer into which the resulting HMAC256 hashed, base64 encoded message will
+ * be placed.
+ * @param[in] ulOutputSize Size of \p pucOutput.
+ * @param[out] pulOutputLength The output length of \p pucOutput.
+ * @return An #AzureIoTResult_t with the result of the operation.
+ */
+AzureIoTResult_t AzureIoT_Base64HMACCalculate( AzureIoTGetHMACFunc_t xAzureIoTHMACFunction,
```
jspaith, line 142: Is this a function we think applications would ever call directly?  If not could we move it to a private header file?

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+typedef uint32_t ( * AzureIoTGetHMACFunc_t )( const uint8_t * pucKey,
+                                              uint32_t ulKeyLength,
+                                              const uint8_t * pucData,
+                                              uint32_t ulDataLength,
+                                              uint8_t * pucOutput,
+                                              uint32_t ulOutputLength,
+                                              uint32_t * pulBytesCopied );
+
+/**
+ * @brief Initialize Azure IoT middleware.
```
jspaith, line 67: I'm assuming this should be called once per process, correct?  (If process is even right word for RTOS, which I think it is even though there's just the 1.)  This singletnon-ness of the API even though seemingly obivous I think worth documenting.

source/include/azure_iot.h
jspaith: And assuming this can move to a private header, related struct/function pointers (e.g. AzureIoTGetHMACFunc_t) could move as well??

jspaith (approved):

source/include/azure_iot.h
danewalton: yup pulling from the embedded SDK

source/include/azure_iot.h
```diff
@@ -0,0 +1,153 @@
...
+ *
+ */
+
+#ifndef AZURE_IOT_H
+#define AZURE_IOT_H
+
+/* AZURE_IOT_DO_NOT_USE_CUSTOM_CONFIG allows building the azure iot library
+ * without a custom config. If a custom config is provided, the
+ * AZURE_IOT_DO_NOT_USE_CUSTOM_CONFIG macro should not be defined. */
+#ifndef AZURE_IOT_DO_NOT_USE_CUSTOM_CONFIG
```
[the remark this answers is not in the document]
danewalton: I mean in this case should it just be
```c
#ifndef AZURE_IOT_CUSTOM_CONFIG
```

and the custom config header which we provide as a baseline has that defined?

source/include/azure_iot.h
danewalton: good call noted

source/include/azure_iot.h
danewalton: I say put it in private and we can always expose if needed. We can't go the other way.

danewalton: I do want to do that in another PR though. There's a greater discussion that needs to happen around directory structure and what to name things. I'll make a note to tackle.

hihigupt (approved):

PR #47, opened 2021-03-24
Add IoT sample for MIMXRT1060

hihigupt:


demo/Common/mimxrt1060/main.c
```diff
@@ -0,0 +1,501 @@
...
+/* MDIO operations. */
+#define mainMDIO_OPS                        enet_ops
+
+/* PHY operations. */
+#define mainPHY_OPS        		            phyksz8081_ops
+
+/* ENET clock frequency. */
+#define mainCLOCK_FREQ                      CLOCK_GetFreq( kCLOCK_IpgClk )
+
+
```
danewalton, line 113: nit nit: uncrustify this at all?

danewalton (approved):
Just to clarify here....this still uses the same task and demo that the other samples use correct? Because I see no calls to any middleware stuff here

demo/Common/mimxrt1060/main.c
hihigupt: Updated

hihigupt:
> Just to clarify here....this still uses the same task and demo that the other samples use correct? Because I see no calls to any middleware stuff here

Reason is because the FreeRTOS task remain same, I am just calling the function that is generic to all the ports.

vStartSimpleMQTTDemo

PR #48, opened 2021-03-24
Add CoreMQTT port

hihigupt:


ports/coreMQTT/azure_iot_core_mqtt.c
```diff
@@ -0,0 +1,233 @@
...
+ */
+ 
+#include <assert.h>
+
+#include "azure_iot_mqtt.h"
+
+/**
+ * Maps CoreMQTT errors to AzureIoTMQTT errors.
+ **/
+static AzureIoTMQTTResult_t prvGetAzureIoTMQTTResult( MQTTStatus_t xResult )
```
jspaith, line 17: Nit (because who doesn't love all these nits :))? - I think instead of Get this should be Translate, because Get usually either implies a direct accessor or some sort of API call or something to retrieve data.

ports/coreMQTT/azure_iot_core_mqtt.c
```diff
@@ -0,0 +1,233 @@
...
+                                        const AzureIoTTransportInterface_t * pxTransportInterface,
+                                        AzureIoTMQTTGetCurrentTimeFunc_t xGetTimeFunction,
+                                        AzureIoTMQTTEventCallback_t xUserCallback,
+                                        uint8_t * pucNetworkBuffer,
+                                        uint32_t ulNetworkBufferLength )
+{
+    MQTTFixedBuffer_t xBuffer = { pucNetworkBuffer, ulNetworkBufferLength };
+    MQTTStatus_t xResult;
+
+    /* Check memory equivalence, but ordering is not guarantee */
```
jspaith, line 109: Nit: "guarantee" => "guaranteed".

ports/coreMQTT/azure_iot_core_mqtt.c
```diff
@@ -0,0 +1,233 @@
...
+    assert( sizeof( AzureIoTMQTTConnectInfo_t ) == sizeof( MQTTConnectInfo_t ) );
```
jspaith, line 110: Is there a compile time, non-hacky, C99 friendly way to check this assert ala C++ `CASSERT`.  All my requirements around this make me think answer is no :), but in perfect world if this was wrong it'd be a compile error.

ports/coreMQTT/azure_iot_core_mqtt.c
```diff
@@ -0,0 +1,233 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/**
+ * @file azure_iot_core_mqtt.c
+ * @brief Implements the port for Azure IoT MQTT based on coreMQTT.
```
jspaith, line 6: I know it's in header but may be worth calling out again that the interface at level is subject to change.

When we get really serious about locking down this level's contract we should link back to some sort of MQTT porting guide doc, but that is in the future at this point.

jspaith (approved):

ports/coreMQTT/azure_iot_core_mqtt.c
ericwol-msft:
```
#if sizeof( AzureIoTMQTTConnectInfo_t ) != sizeof( MQTTConnectInfo_t )
#error Only Debug builds are supported
#endif
```

hihigupt: Discuss offline.

Possible solution can be to define a macro:

#define azureiotSTATIC_ASSERT(COND,MSG) typedef char static_assertion_##MSG[(COND)?1:-1]


but the error generated by this marco, is not useful.

ports/coreMQTT/azure_iot_core_mqtt.c
```diff
@@ -0,0 +1,236 @@
...
+
+    return xReturn;
+}
+
+AzureIoTMQTTResult_t AzureIoTMQTT_Init( AzureIoTMQTTHandle_t xContext,
+                                        const AzureIoTTransportInterface_t * pxTransportInterface,
+                                        AzureIoTMQTTGetCurrentTimeFunc_t xGetTimeFunction,
+                                        AzureIoTMQTTEventCallback_t xUserCallback,
+                                        uint8_t * pucNetworkBuffer,
+                                        uint32_t ulNetworkBufferLength )
```
ericwol-msft, line 107: not size_t?

ericwol-msft (approved):

ports/coreMQTT/azure_iot_core_mqtt.c
hihigupt: will update

ports/coreMQTT/azure_iot_core_mqtt.c
hihigupt: #ignore

PR #49, opened 2021-03-25
Add Provisioning impl

hihigupt:


source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+ *
+ * State transitions :
+ *          CONNECT -> { SUBSCRIBE | COMPLETE } -> { SUBSCRIBING | COMPLETE } -> { REQUEST | COMPLETE } ->
+ *          REQUESTING -> { RESPONSE | COMPLETE } -> { WAITING | COMPLETE } -> { REQUEST | COMPLETE }
+ *
+ * Note : At COMPLETE state we should check the status of lastOperation to see if WorkFlow( WF )
+ *        finshed without error.
+ *
+ **/
+static void prvAzureIoTProvisioningClientUpdateState( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle,
```
jspaith, line 67: Nit/consider: Having to prefix static functions in this file with prvAzureIoTProvisioningClient* is making them pretty long given this isn't public facing.  Any chance we could (well we obviously could:)) / should change these to be shorter?  

The only place that could get confused would be I think on source editors doing function lookup, but even getting down to prvProvClient* would make that unlikely.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+        xMQTTPublishInfo.pcTopicName = xAzureIoTProvisioningClientHandle->_internal.pucProvisioningScratchBuffer;
+        xMQTTPublishInfo.usTopicNameLength = ( uint16_t ) xMQTTTopicLength;
+        xMQTTPublishInfo.pvPayload = xAzureIoTProvisioningClientHandle->_internal.pucProvisioningScratchBuffer + xMQTTTopicLength;
+        xMQTTPublishInfo.xPayloadLength = xMQTTPayloadLength;
+        usPublishPacketIdentifier = AzureIoTMQTT_GetPacketId( &( xAzureIoTProvisioningClientHandle->_internal.xMQTTContext ) );
+
+        if( ( xResult = AzureIoTMQTT_Publish( &( xAzureIoTProvisioningClientHandle->_internal.xMQTTContext ),
+                                              &xMQTTPublishInfo, usPublishPacketIdentifier ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to publish. Error=%d", xResult ) );
```
jspaith, line 353: Nit: Consider only a few extra words here, like `Failed to publish prov request` or something.  My concern would be if we end up with error logs at some point without __FUNCTION__ / __LINE__ / __FILE__ then there's change this could be somewhat ambiguous if there's >1 "failed to publish."

We may be able to piece together where the line we cared about was from context/other log messages, but doing a straight search through src for exact msg is easier.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    }
+}
+/*-----------------------------------------------------------*/
+
+/**
+ *
+ * Implementation of parsing response action, this action is only allowed in azureiotprovisioningWF_STATE_RESPONSE
+ *
+ * */
+static void prvAzureIoTProvisioningClientParseResponse( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle )
```
jspaith, line 376: Another nit around length - does `xAzureIoTProvisioningClientHandle` have to be so long in general?  It's scoped local so no risk of name collisions, and by time we add "_internal" it's getting long.

Again my sense if the fact this isn't a global and the file/names & dirs have "azure" all over the place already goes a long way around making intent clear here.  I'll just float out `xProvClientHandle` maybe??

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    /* Check the state.  */
+    if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_RESPONSE )
+    {
+        xCoreResult =
+            az_iot_provisioning_client_parse_received_topic_and_payload( &xAzureIoTProvisioningClientHandle->_internal.xProvisioningClientCore,
+                                                                         xTopic, xPayload, &xAzureIoTProvisioningClientHandle->_internal.xRegisterResponse );
+
+        if( az_result_failed( xCoreResult ) )
+        {
+            prvAzureIoTProvisioningClientUpdateState( xAzureIoTProvisioningClientHandle, eAzureIoTProvisioningFailed );
```
jspaith, line 393: Nit  - here you call the UpdateState function and then log error but a few lines later you call the log error and then the update state in different order.

I'd recommend logging 1st though I don't care that much beyond consistency.  The "Log/DosomeError" is how I've seen this historically, and on the case that something weird happens in the function called in the else{errorBlock}, then there's at least a chance that the log error will be output.  (Pretending stdout gets flushed out of course:)).

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+            return;
+        }
+
+        if( xAzureIoTProvisioningClientHandle->_internal.xRegisterResponse.operation_status == AZ_IOT_PROVISIONING_STATUS_ASSIGNED )
+        {
+            prvAzureIoTProvisioningClientUpdateState( xAzureIoTProvisioningClientHandle, eAzureIoTProvisioningSuccess );
+        }
+        else if( xAzureIoTProvisioningClientHandle->_internal.xRegisterResponse.retry_after_seconds == 0 )
+        {
+            AZLogError( ( "IoTProvisioning client failed with error %d, extended error status: %u\r\n",
```
jspaith, line 404: Nit - i'd add a fact that server has 0 for retry so it's clear why we're giving up.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+                          xAzureIoTProvisioningClientHandle->_internal.xRegisterResponse.registration_state.error_code,
+                          xAzureIoTProvisioningClientHandle->_internal.xRegisterResponse.registration_state.extended_error_code ) );
+            /* Server responded with error with no retry.  */
+            prvAzureIoTProvisioningClientUpdateState( xAzureIoTProvisioningClientHandle, eAzureIoTProvisioningServerError );
+        }
+        else
+        {
+            xAzureIoTProvisioningClientHandle->_internal.ullProvisioningRequestTimeout =
```
jspaith, line 412: Nit - again consider a Log about this state change, this one feels particualryl important.  Now iif the log ends up in the `prvAzureIoTProvisioningClientUpdateState` (as I contemplated kind of above) then don't need one here.

jspaith, line 412: Is `ullProvisioningRequestTimeout` the right name for this?  It seems like the absolute time since you're doing a xGetTime() and adding the delta.  But this feels more like a `ulWhenToRetry` or something?

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+/**
+ *
+ * Implementation of parsing response action, this action is only allowed in azureiotprovisioningWF_STATE_RESPONSE
+ *
+ * */
+static void prvAzureIoTProvisioningClientParseResponse( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle )
+{
+    az_result xCoreResult;
+    az_span xPayload = az_span_create( xAzureIoTProvisioningClientHandle->_internal.pucProvisioningLastResponsePayload,
+                                       ( int32_t ) xAzureIoTProvisioningClientHandle->_internal.xProvisioningLastResponsePayloadLength );
```
jspaith, line 380: Another nit around length - sorry - but again `xProvisioningLastResponsePayloadLength ` does that need to be so long?  The fact it's internal to a handle about provisioning makes me think this borders on repeating ourselves?

jspaith (commented):
@hihigupt looking good so far.  I need to drop offline and I'm only about 1/3 way through, but sharing what I have to give you a head start if you want.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+
+    if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_SUBSCRIBE )
+    {
+        xMQTTSubscription.xQoS = eAzureIoTMQTTQoS0;
+        xMQTTSubscription.pcTopicFilter = ( const uint8_t * ) AZ_IOT_PROVISIONING_CLIENT_REGISTER_SUBSCRIBE_TOPIC;
+        xMQTTSubscription.usTopicFilterLength = ( uint16_t ) sizeof( AZ_IOT_PROVISIONING_CLIENT_REGISTER_SUBSCRIBE_TOPIC ) - 1;
+
+        usSubscribePacketIdentifier = AzureIoTMQTT_GetPacketId( &( xAzureIoTProvisioningClientHandle->_internal.xMQTTContext ) );
+
+        AZLogInfo( ( "Attempt to subscribe to the MQTT  %s topics.\r\n", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
```
ericwol-msft, line 229:
```suggestion
        AZLogInfo( ( "AzureIoTProvisioning attempting to subscribe to the MQTT %s topic.\r\n", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+
+        if( ( xResult = AzureIoTMQTT_Subscribe( &( xAzureIoTProvisioningClientHandle->_internal.xMQTTContext ),
+                                                &xMQTTSubscription, 1, usSubscribePacketIdentifier ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to SUBSCRIBE to MQTT topic. Error=%d", xResult ) );
```
ericwol-msft, line 234:
```suggestion
            AZLogError( ( "AzureIoTProvisioning failed to SUBSCRIBE to MQTT topic (%d).", xResult ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+        else
+        {
+            xReturn = eAzureIoTProvisioningSuccess;
+        }
+
+        prvAzureIoTProvisioningClientUpdateState( xAzureIoTProvisioningClientHandle, xReturn );
+    }
+    else
+    {
+        AZLogWarn( ( "Subscribe action called in wrong state %d\r\n",
```
ericwol-msft, line 246:
```suggestion
        AZLogWarn( ( "AzureIoTProvisioning subscribe action called in wrong state (%d)\r\n",
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    if( az_result_failed( az_json_writer_init( &xJsonWriter, xBuffer, NULL ) ) ||
+        az_result_failed( az_json_writer_append_begin_object( &xJsonWriter ) ) ||
+        az_result_failed( az_json_writer_append_property_name( &xJsonWriter, xRegistrationIdLabel ) ) ||
+        az_result_failed( az_json_writer_append_string( &xJsonWriter, xRegistrationId ) ) ||
+        ( ( xAzureIoTProvisioningClientHandle->_internal.pucProvisioningRegistrationPayload != NULL ) &&
+          ( az_result_failed( az_json_writer_append_property_name( &xJsonWriter, xPayloadLabel ) ) ||
+            az_result_failed( az_json_writer_append_json_text( &xJsonWriter, xPayload ) ) ) ) ||
+        az_result_failed( az_json_writer_append_end_object( &xJsonWriter ) ) )
+    {
+        AZLogError( ( "Failed to create Request JSON payload." ) );
```
ericwol-msft, line 276:
```suggestion
        AZLogError( ( "AzureIoTProvisioning failed to create Request JSON payload." ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+
+    *pulBytesUsed = ( uint32_t ) az_span_size( az_json_writer_get_bytes_used_in_destination( &xJsonWriter ) );
+
+    return xResult;
+}
+/*-----------------------------------------------------------*/
+
+/**
+ *
+ * Implementation of request action, this action is only allowed in azureiotprovisioningWF_STATE_REQUEST
```
ericwol-msft, line 292:
```suggestion
 * Implementation of request action. This action is only allowed in azureiotprovisioningWF_STATE_REQUEST
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    #define azureiotprovisioningCONNACK_RECV_TIMEOUT_MS    azureiotCONNACK_RECV_TIMEOUT_MS
+#endif /* azureiotprovisioningCONNACK_RECV_TIMEOUT_MS */
+
+#ifndef azureiotprovisioningUSER_AGENT
+    #define azureiotprovisioningUSER_AGENT             ""
+#endif /* azureiotprovisioningUSER_AGENT */
+
+#define azureiotprovisioningPROCESS_LOOP_TIMEOUT_MS    ( 500U )
+
+#define azureiotprovisioningWF_STATE_INIT              ( 0x0 )
```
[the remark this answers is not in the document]
hihigupt: Then we have to expose internal type in header

---
In reply to: [602643440](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602643440) [](ancestors = 602643440)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    AzureIoTProvisioningClientResult_t xReturn;
+    AzureIoTMQTTResult_t xResult;
+    uint8_t ucSessionPresent;
+    uint32_t ulBytesCopied = 0;
+    size_t xMQTTUsernameLength;
+    az_result xCoreResult;
+
+    if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState != azureiotprovisioningWF_STATE_CONNECT )
+    {
+        AZLogWarn( ( "Connect action called in wrong state %d\r\n",
```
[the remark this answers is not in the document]
hihigupt: Only reason is the COM port client (PUTTY). Because they generally have some configuration to \n -> newline.

---
In reply to: [602576961](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602576961) [](ancestors = 602576961)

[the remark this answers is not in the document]
hihigupt: I will remove the \r\n from logs and let the log implementation to add new line.

---
In reply to: [603467682](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603467682) [](ancestors = 603467682,602576961)

source/azure_iot_provisioning_client.c
hihigupt: xProvClientHandle 

---
In reply to: [602632111](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602632111) [](ancestors = 602632111)

source/azure_iot_provisioning_client.c
hihigupt: I can add Debug level log in prvAzureIoTProvisioningClientUpdateState to record the state changes

---
In reply to: [602632933](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602632933) [](ancestors = 602632933)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+        }
+    }
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * Trigger state machine action base on the state.
+ *
+ **/
+static void prvTriggerAction( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle )
```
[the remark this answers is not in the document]
hihigupt: Make sense. I was following kind of azrtos function name to always prefix all the function with the file name, which helps in gdb (using tab).

---
In reply to: [603459560](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603459560) [](ancestors = 603459560)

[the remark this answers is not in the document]
hihigupt: I have updated the names, let me know your thoughts 

---
In reply to: [603512044](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603512044) [](ancestors = 603512044,603459560)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+static void prvMQTTProcessSubAck( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle,
+                                  AzureIoTMQTTPublishInfo_t * pxPacketInfo )
+{
+    ( void ) pxPacketInfo;
+
+    prvAzureIoTProvisioningClientUpdateState( xAzureIoTProvisioningClientHandle, eAzureIoTProvisioningSuccess );
```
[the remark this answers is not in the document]
hihigupt: I thought about it, but to make things simpler, this implementation took dependency on how service is implemented. Any failure of this kind will lead to TCP close, which will lead to MQTT layer errors. 

By doing the extra parsing, will help us better insight on how the workflow failed, but with adding code size cost.


---
In reply to: [603463747](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603463747) [](ancestors = 603463747)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+/**
+ * Process MQTT Response from Provisioning Service
+ *
+ */
+static void prvMQTTProcessResponse( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle,
+                                    AzureIoTMQTTPublishInfo_t * pxPublishInfo )
+{
+    if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_REQUESTING )
+    {
+        if( ( pxPublishInfo->xPayloadLength + pxPublishInfo->usTopicNameLength ) <=
```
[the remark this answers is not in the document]
hihigupt: pxPublishInfo->xPayloadLength + pxPublishInfo->usTopicNameLength is already accommodating for topic and payload length. Am I missing something here?

---
In reply to: [603484273](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603484273) [](ancestors = 603484273)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle = ( AzureIoTProvisioningClientHandle_t ) pxMQTTContext;
+
+    if( ( pxPacketInfo->ucType & 0xF0U ) == azureiotmqttPACKET_TYPE_SUBACK )
+    {
+        prvMQTTProcessSubAck( xAzureIoTProvisioningClientHandle, pxDeserializedInfo->pxPublishInfo );
+    }
+    else if( ( pxPacketInfo->ucType & 0xF0U ) == azureiotmqttPACKET_TYPE_PUBLISH )
+    {
+        prvMQTTProcessResponse( xAzureIoTProvisioningClientHandle, pxDeserializedInfo->pxPublishInfo );
+    }
```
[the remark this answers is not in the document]
hihigupt: there will be others like Ping Response packets. How about adding info level logs, rather then warning?

---
In reply to: [603488607](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603488607) [](ancestors = 603488607)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    ulBufferLeft = ulSasBufferLen - ulBytesUsed;
+
+    if( ulBufferLeft < ulHMACBufferLength )
+    {
+        AZLogError( ( "IoTProvisioning token generation failed with out of memory" ) );
+        return eAzureIoTProvisioningOutOfMemory;
+    }
+
+    ulBufferLeft -= ulHMACBufferLength;
+    pucHMACBuffer = pucSASBuffer + ulSasBufferLen - ulHMACBufferLength;
```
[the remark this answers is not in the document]
hihigupt: I am calculating the result in the buffer that is passed to prvAzureIoTProvisioningClientTokenGet. Now if I am calculating things that are temporary at the end of buffer, then I do not have to move the result back to start of buffer.

---
In reply to: [603492732](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603492732) [](ancestors = 603492732)

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
jspaith: Indeed, you're checking the topic length above.  But I'm not seeing a corresponding check for whether the `pxPublishInfo->xPayloadLength` is too big for the `xAzureIoTProvisioningClientHandle->_internal.pucProvisioningLastResponsePayload` buffer.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+ *          REQUESTING -> { RESPONSE | COMPLETE } -> { WAITING | COMPLETE } -> { REQUEST | COMPLETE }
+ *
+ * Note : At COMPLETE state we should check the status of lastOperation to see if WorkFlow( WF )
+ *        finshed without error.
+ *
+ **/
+static void prvAzureIoTProvisioningClientUpdateState( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle,
+                                                      uint32_t ulActionResult )
+{
+    uint32_t ulState = xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState;
```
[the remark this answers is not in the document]
hihigupt: Added log statement

---
In reply to: [602566005](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602566005) [](ancestors = 602566005)

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
jspaith: Ahh - Info or even debug makes more sense I'd forgotten about PING resp.

source/azure_iot_provisioning_client.c
hihigupt: I renamed all with prefix prvProvClient

---
In reply to: [602629054](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602629054) [](ancestors = 602629054)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    }
+    else
+    {
+        memset( xAzureIoTProvisioningClientHandle, 0, sizeof( AzureIoTProvisioningClient_t ) );
+        /* Setup scratch buffer to be used by middleware */
+        xAzureIoTProvisioningClientHandle->_internal.ulProvisioningScratchBufferLength =
+            azureMAX( ( azureiotUSERNAME_MAX + azureiotPASSWORD_MAX ),
+                      ( azureiotTOPIC_MAX + azureiotPROVISIONING_REQUEST_PAYLOAD_MAX ) );
+        xAzureIoTProvisioningClientHandle->_internal.pucProvisioningScratchBuffer = pucBuffer;
+        pucBuffer += xAzureIoTProvisioningClientHandle->_internal.ulProvisioningScratchBufferLength;
```
jspaith, line 745: Nit - consider a separate variable name for this, instead of re-using passed in parameter (compiler optimizers should reduce out in any case).  I think that makes a bit easier to read, like `pucNetworkBuffer` or something?

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+{
+    AzureIoTMQTTResult_t xResult;
+    AzureIoTProvisioningClientResult_t xReturn;
+    uint32_t ulWaitTime;
+
+    do
+    {
+        if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_COMPLETE )
+        {
+            break;
```
jspaith, line 501: We should log out `if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_COMPLETE )` case, since the way the API works & state machine someone may not realize fatal errors or even just re-running provisionion request would need a reset of handle.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    }
+}
+/*-----------------------------------------------------------*/
+
+/**
+ *
+ * Implementation of wait check action, this action is only allowed in azureiotprovisioningWF_STATE_WAITING
+ *
+ * */
+static void prvAzureIoTProvisioningClientCheckTimeout( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle )
```
jspaith, line 430: 1 -- 
What happens if (a) our provisioning request "fails" but the server sends a "try again in 20 minutes" or something kind of long?  It naively seems to me that we will stay in the WAITING state of the state machine - which seems reasonable, except I would worry that we'll never invoke the underlying MQTT layer to do an MQTT PING and the connection could end up being dropped on us.

Is this a reasonable read / concern?

2 -- Should we have a mechanism to return `ullProvisioningRequestTimeout` back to the application?  For instance if Hub says "I'll be ready in 10 hours" that could impact a sophisticated application's desire to retry.  Though I don't know if DPS would return such a long delay in 1st place or error out?

FWIW there's no harm in NOT having a `ullProvisioningRequestTimeout` returner, since we can always add later, but want to at least see your thinking on it?

source/azure_iot_provisioning_client.c
hihigupt: ullProvisioningRetryAfter

---
In reply to: [602633333](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r602633333) [](ancestors = 602633333)

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
hihigupt: Ah... Currently there is only one buffer called ucProvisioningLastResponse. In which we copy all the topic + payload. Now to see where the topic and payload start we have auxillary pointers (pucProvisioningLastResponseTopic,pucProvisioningLastResponsePayload)

---
In reply to: [603532765](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603532765) [](ancestors = 603532765)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+    }
+    else if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState != azureiotprovisioningWF_STATE_COMPLETE )
+    {
+        AZLogError( ( "Provisioning client state is not in complete state" ) );
+        xReturn = eAzureIoTProvisioningFailed;
+    }
+    else
+    {
+        *pulExtendedErrorCode =
+            xAzureIoTProvisioningClientHandle->_internal.xRegisterResponse.registration_state.extended_error_code;
```
jspaith, line 929: 1 - 
One minor concern I have here is that we can get into STATE_COMPLETE where we think we have the extended error code but in reality failed for some other reason.

I initially thought we could just check `( xAzureIoTProvisioningClientHandle->_internal.lastOperationResult )` in this function before returning `eAzureIoTProvisioningSuccess`.  I don't think we can set the lastOperationResult to some sort of failure *AND* get a DPS result code here?

But that leads to 

2 -- Can we even check on `azureiotprovisioningWF_STATE_COMPLETE` here?  It seems naively that we could get an error from hub saying "Call back again in 2 minutes" and we wouldn't be in COMPLETE, but I think the extended error code would still be set?  This could potentially be meaningful about extnded errors about why I'm being told to call back later so maybe this should be plumbed (assuming DPS even would do a call-back-later and extended code at same time??)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+        xAzureIoTProvisioningClientHandle->_internal.pxProvisioningTokenRefresh = prvAzureIoTProvisioningClientTokenGet;
+        xAzureIoTProvisioningClientHandle->_internal.xProvisioningHMACFunction = xHmacFunction;
+        xReturn = eAzureIoTProvisioningSuccess;
+    }
+
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTProvisioningClientResult_t AzureIoTProvisioningClient_Register( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle,
```
jspaith, line 833: Not directly related to this but more doxygen for `AzureIoTProvisioningClient_Register`, but... I think we need to be really clear about what a StatusSuccess means here.

It means that the Hub has acknowledged our request and NOT given us some sort of pending code.  But it does NOT mean that we're happy campers, because in the SDK we don't parse out 200/300/400/500 HTTP status codes.  So doxygen saying `After a successful result, AzureIoTProvisioningClient_HubGet() can be called to get the IoT Hub and device ID.`, I really think the flow needs to be like (with actual word-smithing)

"After successful result, your app needs to call AzureIoTProvisioningClient_GetExtendedCode and if it has a happy (200-299??) HTTP status code for top 3 digits, then and only then can it call GetHub().  Otherwise the register call has completed but not succeeded in terms of having device registered"

We could also potentially have the SDK have opinions and not return "statusSuccess" if a 500 came from hub with no retry guidance?  That would be a more natural way I'd think API would work fwiw.

jspaith: And in any case note that [here](https://github.com/Azure/azure-iot-middleware-freertos/blob/f5e07fba53698f26ce03d172ce51134395ad8d59/source/include/azure_iot_provisioning_client.h#L153) the doxygen is calling `AzureIoTProvisioningClient_HubGet` but really should be `AzureIoTProvisioningClient_GetHub`.

jspaith (commented):
I've made it to the end although I'd like to sleep on this some before formal sign-off and give you chance to look at any feedback.

In general this looks really good and appreciate specifically use of state machine to drive all this.

source/azure_iot_provisioning_client.c
hihigupt: Currently with each action trigger, we also call ProcessLoop of MQTT, so that should take care of PingReq/PingRes.

For 2, I would suggest to add it latter.


---
In reply to: [603546950](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603546950) [](ancestors = 603546950)

source/azure_iot_provisioning_client.c
hihigupt: Ack.

---
In reply to: [603542867](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603542867) [](ancestors = 603542867)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+ */
+static void prvMQTTProcessResponse( AzureIoTProvisioningClientHandle_t xAzureIoTProvisioningClientHandle,
+                                    AzureIoTMQTTPublishInfo_t * pxPublishInfo )
+{
+    if( xAzureIoTProvisioningClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_REQUESTING )
+    {
+        if( ( pxPublishInfo->xPayloadLength + pxPublishInfo->usTopicNameLength ) <=
+            sizeof( xAzureIoTProvisioningClientHandle->_internal.ucProvisioningLastResponse ) )
+        {
+            xAzureIoTProvisioningClientHandle->_internal.pucProvisioningLastResponsePayload =
```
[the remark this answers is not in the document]
hihigupt: Removed

---
In reply to: [603482357](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603482357) [](ancestors = 603482357)

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
hihigupt: Update.

---
In reply to: [603550398](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603550398) [](ancestors = 603550398,603532765)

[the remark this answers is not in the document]
hihigupt: Let me know if change address your concern.

---
In reply to: [603522527](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603522527) [](ancestors = 603522527,603484273)

source/azure_iot_provisioning_client.c
hihigupt: Regarding 2: We are checking for complete state : line 921.

---
In reply to: [603557149](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603557149) [](ancestors = 603557149)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+    {
+        AZLogError( ( "AzureIoTProvisioning failed to create Request JSON payload." ) );
+        xResult = eAzureIoTProvisioningFailed;
+    }
+    else
+    {
+        xResult = eAzureIoTProvisioningSuccess;
+    }
+
+    *pulBytesUsed = ( uint32_t ) az_span_size( az_json_writer_get_bytes_used_in_destination( &xJsonWriter ) );
```
ericwol-msft, line 292: do this even if xResult failed? az_json_writer_init could have failed.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+    AzureIoTMQTTPublishInfo_t xMQTTPublishInfo = { 0 };
+    size_t xMQTTTopicLength;
+    uint32_t xMQTTPayloadLength = 0;
+    uint16_t usPublishPacketIdentifier;
+    az_result xCoreResult;
+
+    /* Check the state.  */
+    if( xProvClientHandle->_internal.ulWorkflowState != azureiotprovisioningWF_STATE_REQUEST )
+    {
+        AZLogWarn( ( "Request action called in wrong state %d",
```
ericwol-msft, line 316:
```suggestion
        AZLogWarn( ( "AzureIoTProvisioning client request action called in wrong state (%d).",
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+                                               xMQTTTopicLength,
+                                               xProvClientHandle->_internal.ulScratchBufferLength -
+                                               ( uint32_t ) xMQTTTopicLength,
+                                               &xMQTTPayloadLength ) ) != eAzureIoTProvisioningSuccess )
+        {
+            prvProvClientUpdateState( xProvClientHandle, xResult );
+            return;
+        }
+
+        ( void ) memset( ( void * ) &xMQTTPublishInfo, 0x00, sizeof( xMQTTPublishInfo ) );
```
ericwol-msft, line 356: seams redundant with line 308

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+        else
+        {
+            ulWaitTime = ulTimeoutMilliseconds;
+            ulTimeoutMilliseconds = 0;
+        }
+
+        prvTriggerAction( xAzureIoTProvisioningClientHandle );
+
+        if( ( xResult =
+                  AzureIoTMQTT_ProcessLoop( &( xAzureIoTProvisioningClientHandle->_internal.xMQTTContext ),
```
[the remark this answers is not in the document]
hihigupt: That will break the contract in header [https://github.com/Azure/azure-iot-middleware-freertos/blob/f5e07fba53698f26ce03d172ce51134395ad8d59/source/interface/azure_iot_mqtt.h#L329](https://github.com/Azure/azure-iot-middleware-freertos/blob/f5e07fba53698f26ce03d172ce51134395ad8d59/source/interface/azure_iot_mqtt.h#L329).


Atleast in case of coreMQTT this will be a duplicate code at middleware layer.

---
In reply to: [603458639](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603458639) [](ancestors = 603458639)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+        xMQTTPublishInfo.pcTopicName = xProvClientHandle->_internal.pucScratchBuffer;
+        xMQTTPublishInfo.usTopicNameLength = ( uint16_t ) xMQTTTopicLength;
+        xMQTTPublishInfo.pvPayload = xProvClientHandle->_internal.pucScratchBuffer + xMQTTTopicLength;
+        xMQTTPublishInfo.xPayloadLength = xMQTTPayloadLength;
+        usPublishPacketIdentifier = AzureIoTMQTT_GetPacketId( &( xProvClientHandle->_internal.xMQTTContext ) );
+
+        if( ( xMQTTResult = AzureIoTMQTT_Publish( &( xProvClientHandle->_internal.xMQTTContext ),
+                                                  &xMQTTPublishInfo, usPublishPacketIdentifier ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to publish prov request, error=%d", xMQTTResult ) );
```
ericwol-msft, line 367:
```suggestion
            AZLogError( ( "AzureIoTProvisioning failed to publish provisioning request (%d)", xMQTTResult ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+    az_span xTopic = az_span_create( xProvClientHandle->_internal.ucProvisioningLastResponse,
+                                     ( int32_t ) xProvClientHandle->_internal.usLastResponseTopicLength );
+    az_span xPayload = az_span_create( xProvClientHandle->_internal.ucProvisioningLastResponse +
+                                            xProvClientHandle->_internal.usLastResponseTopicLength,
+                                       ( int32_t ) xProvClientHandle->_internal.xLastResponsePayloadLength );
+
+    /* Check the state.  */
+    if( xProvClientHandle->_internal.ulWorkflowState != azureiotprovisioningWF_STATE_RESPONSE )
+    {
+        AZLogWarn( ( "AzureIoTProvisioning Parse response action called in wrong state %d",
```
ericwol-msft, line 397:
```suggestion
        AZLogWarn( ( "AzureIoTProvisioning parse response action called in wrong state (%d)",
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+            return;
+        }
+
+        if( xProvClientHandle->_internal.xRegisterResponse.operation_status == AZ_IOT_PROVISIONING_STATUS_ASSIGNED )
+        {
+            prvProvClientUpdateState( xProvClientHandle, eAzureIoTProvisioningSuccess );
+        }
+        else if( xProvClientHandle->_internal.xRegisterResponse.retry_after_seconds == 0 )
+        {
+            AZLogError( ( "AzureIoTProvisioning client failed with error %d, extended error status: %u and server retry time 0",
```
ericwol-msft, line 419:
```suggestion
            AZLogError( ( "AzureIoTProvisioning client failed with error %d, extended error status: %u and no server retry time duration",
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+        }
+        else
+        {
+            xProvClientHandle->_internal.ullRetryAfter =
+                xProvClientHandle->_internal.xGetTimeFunction() +
+                    xProvClientHandle->_internal.xRegisterResponse.retry_after_seconds;
+            prvProvClientUpdateState( xProvClientHandle, eAzureIoTProvisioningPending );
+        }
+    }
+
```
ericwol-msft: nit, extra line

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+            prvProvClientUpdateState( xProvClientHandle, eAzureIoTProvisioningPending );
+        }
+    }
+    
+}
+/*-----------------------------------------------------------*/
+
+/**
+ *
+ * Implementation of wait check action, this action is only allowed in azureiotprovisioningWF_STATE_WAITING
```
ericwol-msft, line 438:
```suggestion
 * Implementation of wait check action. This action is only allowed in azureiotprovisioningWF_STATE_WAITING
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+        case azureiotprovisioningWF_STATE_COMPLETE:
+            /* None action taken here, as these states are waiting for receive path. */
+            break;
+
+        case azureiotprovisioningWF_STATE_WAITING:
+            prvProvClientCheckTimeout( xProvClientHandle );
+            break;
+
+        default:
+            AZLogError( ( "State not handled %d",
```
ericwol-msft, line 490:
```suggestion
            AZLogError( ( "AzureIoTProvisioning state not handled (%d)",
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+            ulTimeoutMilliseconds = 0;
+        }
+
+        prvProvClientTriggerAction( xProvClientHandle );
+
+        if( ( xResult =
+                  AzureIoTMQTT_ProcessLoop( &( xProvClientHandle->_internal.xMQTTContext ),
+                                            ulWaitTime ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to process loop: ProcessLoopDuration=%u, Error=%d", ulTimeoutMilliseconds, xResult ) );
```
ericwol-msft, line 533:
```suggestion
            AZLogError( ( "AzureIoTProvisioning failed to process loop: ProcessLoopDuration=%u, Error=%d", ulTimeoutMilliseconds, xResult ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+                pxPublishInfo->xPayloadLength;
+            memcpy( xProvClientHandle->_internal.ucProvisioningLastResponse +
+                        xProvClientHandle->_internal.usLastResponseTopicLength,
+                    pxPublishInfo->pvPayload, pxPublishInfo->xPayloadLength );
+
+            prvProvClientUpdateState( xProvClientHandle, eAzureIoTProvisioningSuccess );
+        }
+        else
+        {
+            AZLogError( ( "AzureIoTProvisioning process response failed to allocated %d bytes",
```
ericwol-msft, line 595: not really an alloc. buffer too small?

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+                          pxPublishInfo->xPayloadLength + pxPublishInfo->usTopicNameLength ) );
+            prvProvClientUpdateState( xProvClientHandle, eAzureIoTProvisioningOutOfMemory );
+        }
+    }
+}
+/*-----------------------------------------------------------*/
+
+/**
+ * Callback that receives all the Event from MQTT
```
ericwol-msft:
```suggestion
 * Callback that handles all the MQTT events 
```
?

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+                                          pxDeserializedInfo->pxPublishInfo );
+    }
+    else if( ( pxPacketInfo->ucType & 0xF0U ) == azureiotmqttPACKET_TYPE_PUBLISH )
+    {
+        prvProvClientMQTTProcessResponse( xProvClientHandle,
+                                            pxDeserializedInfo->pxPublishInfo );
+    }
+    else
+    {
+        AZLogDebug( ( "AzureIoTProvisioning received packet of type : %d", pxPacketInfo->ucType ) );
```
ericwol-msft, line 625:
```suggestion
        AZLogDebug( ( "AzureIoTProvisioning received packet of type: %d", pxPacketInfo->ucType ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,963 @@
...
+#ifndef azureiotprovisioningKEEP_ALIVE_TIMEOUT_SECONDS
+    #define azureiotprovisioningKEEP_ALIVE_TIMEOUT_SECONDS    azureiotKEEP_ALIVE_TIMEOUT_SECONDS
+#endif /* azureiotprovisioningKEEP_ALIVE_TIMEOUT_SECONDS */
+
+#ifndef azureiotprovisioningCONNACK_RECV_TIMEOUT_MS
+    #define azureiotprovisioningCONNACK_RECV_TIMEOUT_MS    azureiotCONNACK_RECV_TIMEOUT_MS
+#endif /* azureiotprovisioningCONNACK_RECV_TIMEOUT_MS */
+
+#ifndef azureiotprovisioningUSER_AGENT
+    #define azureiotprovisioningUSER_AGENT             ""
```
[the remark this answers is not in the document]
hihigupt: this is similar to embedded sdk default

[https://github.com/Azure/azure-sdk-for-c/blob/fb5d5610c2327afbe611a97d6970964ccbfa0778/sdk/src/azure/iot/az_iot_provisioning_client.c#L46](https://github.com/Azure/azure-sdk-for-c/blob/fb5d5610c2327afbe611a97d6970964ccbfa0778/sdk/src/azure/iot/az_iot_provisioning_client.c#L46)

---
In reply to: [603497120](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603497120) [](ancestors = 603497120)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+    uint32_t ulBytesUsed;
+    uint32_t ulBufferLeft;
+    size_t xLength;
+
+    xCoreResult = az_iot_provisioning_client_sas_get_signature( &( xProvClientHandle->_internal.xProvisioningClientCore ),
+                                                                ullExpiryTimeSecs, xSpan, &xSpan );
+
+    if( az_result_failed( xCoreResult ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning failed failed to get signature with error status: %08x", xCoreResult ) );
```
ericwol-msft, line 653:
```suggestion
        AZLogError( ( "AzureIoTProvisioning failed to get signature with error status: %08x", xCoreResult ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+        return eAzureIoTProvisioningFailed;
+    }
+
+    ulBytesUsed = ( uint32_t ) az_span_size( xSpan );
+    ulBufferLeft = ulSasBufferLen - ulBytesUsed;
+
+    if( ulBufferLeft < ulHMACBufferLength )
+    {
+        AZLogError( ( "AzureIoTProvisioning token generation failed with out of memory" ) );
```
ericwol-msft, line 662:
```suggestion
        AZLogError( ( "AzureIoTProvisioning token generation failed with insufficient buffer size." ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+    }
+
+    ulBufferLeft -= ulHMACBufferLength;
+    pucHMACBuffer = pucSASBuffer + ulSasBufferLen - ulHMACBufferLength;
+
+    if( AzureIoT_Base64HMACCalculate( xProvClientHandle->_internal.xHMACFunction,
+                                      ucKey, ulKeyLen, pucSASBuffer, ulBytesUsed, pucSASBuffer + ulBytesUsed, ulBufferLeft,
+                                      pucHMACBuffer, ulHMACBufferLength, &ulBytesUsed ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning failed to encoded hash" ) );
```
ericwol-msft, line 673:
```suggestion
        AZLogError( ( "AzureIoTProvisioning failed to encoded HMAC hash" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTProvisioningClientResult_t AzureIoTProvisioningClient_OptionsInit( AzureIoTProvisioningClientOptions_t * pxProvisioningClientOptions )
+{
+    AzureIoTProvisioningClientResult_t xReturn;
+
+    if( pxProvisioningClientOptions == NULL )
+    {
+        AZLogError( ( "AzureIoTProvisioning failed to initialize options: Invalid argument " ) );
```
ericwol-msft, line 719:
```suggestion
        AZLogError( ( "AzureIoTProvisioning failed to initialize options: Invalid argument" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,996 @@
...
+        ( pucBuffer == NULL ) || ( ulBufferLength == 0 ) ||
+        ( xGetTimeFunction == NULL ) || ( pxTransportInterface == NULL ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning initialization failed: Invalid argument" ) );
+        xReturn = eAzureIoTProvisioningInvalidArgument;
+    }
+    else if( ( ulBufferLength < ( azureiotTOPIC_MAX + azureiotPROVISIONING_REQUEST_PAYLOAD_MAX ) ) ||
+             ( ulBufferLength < ( azureiotUSERNAME_MAX + azureiotPASSWORD_MAX ) ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning initialize failed: Not enough memory passed " ) );
```
ericwol-msft, line 769:
```suggestion
        AZLogError( ( "AzureIoTProvisioning initialize failed: Insufficient buffer size " ) );
```

source/azure_iot_provisioning_client.c
hihigupt: Currently if the registration on DPS side does not end to Assigned state and there is no retry, we move client side state to ServerError :

[https://github.com/Azure/azure-iot-middleware-freertos/blob/0a198e3f94adb2f557ddca9118cd0692cd0b87bf/source/azure_iot_provisioning_client.c#L424](https://github.com/Azure/azure-iot-middleware-freertos/blob/0a198e3f94adb2f557ddca9118cd0692cd0b87bf/source/azure_iot_provisioning_client.c#L424)

In that case return of this function should give success only in assigned state of DPS.

---
In reply to: [603559700](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603559700) [](ancestors = 603559700)

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+        {
+            AZLogError( ( "AzureIoTProvisioning initialization failed: with core error : %08x", xCoreResult ) );
+            xReturn = eAzureIoTProvisioningFailed;
+        }
+        else if( ( xResult = AzureIoTMQTT_Init( &( xProvClientHandle->_internal.xMQTTContext ),
+                                                pxTransportInterface, prvProvClientGetTimeMillseconds,
+                                                prvProvClientEventCallback, pucNetworkBuffer,
+                                                ulNetworkBufferLength ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "AzureIoTProvisioning initialization failed: with mqtt init error : %d", xResult ) );
```
ericwol-msft, line 815:
```suggestion
            AZLogError( ( "AzureIoTProvisioning initialization failed with mqtt error : %d", xResult ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+void AzureIoTProvisioningClient_Deinit( AzureIoTProvisioningClientHandle_t xProvClientHandle )
+{
+    if( xProvClientHandle == NULL )
+    {
+        AZLogError( ( "AzureIoTProvisioningClient_Deinit failed: Invalid argument" ) );
```
ericwol-msft, line 832:
```suggestion
        AZLogError( ( "AzureIoTProvisioningClient Deinit failed: Invalid argument" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+                                                                               uint32_t ulSymmetricKeyLength,
+                                                                               AzureIoTGetHMACFunc_t xHmacFunction )
+{
+    AzureIoTProvisioningClientResult_t xReturn;
+
+    if( ( xProvClientHandle == NULL ) ||
+        ( pucSymmetricKey == NULL ) || ( ulSymmetricKeyLength == 0 ) ||
+        ( xHmacFunction == NULL ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning client symmetric key fail: Invalid argument" ) );
```
ericwol-msft, line 848:
```suggestion
        AZLogError( ( "AzureIoTProvisioning client symmetric key failed: Invalid argument" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+    uint32_t ulHostnameLength;
+    uint32_t ulDeviceIDLength;
+    AzureIoTProvisioningClientResult_t xReturn;
+    az_span * pxHostname;
+    az_span * pxDeviceID;
+
+    if( ( xProvClientHandle == NULL ) || ( pucHubHostname == NULL ) ||
+        ( pulHostnameLength == NULL ) || ( pucDeviceID == NULL ) || ( pulDeviceIDLength == NULL ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning hub get failed: Invalid argument" ) );
```
ericwol-msft, line 903:
```suggestion
        AZLogError( ( "AzureIoTProvisioning get hub failed: Invalid argument" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+    else
+    {
+        pxHostname = &xProvClientHandle->_internal.xRegisterResponse.registration_state.assigned_hub_hostname;
+        pxDeviceID = &xProvClientHandle->_internal.xRegisterResponse.registration_state.device_id;
+        ulHostnameLength = ( uint32_t ) az_span_size( *pxHostname );
+        ulDeviceIDLength = ( uint32_t ) az_span_size( *pxDeviceID );
+
+        if( ( *pulHostnameLength < ulHostnameLength ) || ( *pulDeviceIDLength < ulDeviceIDLength ) )
+        {
+            AZLogWarn( ( "AzureIoTProvisioning Memory passed is not enough to store hub info " ) );
```
ericwol-msft, line 924:
```suggestion
            AZLogWarn( ( "AzureIoTProvisioning Memory buffer passed is not enough to store hub info" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+
+AzureIoTProvisioningClientResult_t AzureIoTProvisioningClient_GetExtendedCode( AzureIoTProvisioningClientHandle_t xProvClientHandle,
+                                                                               uint32_t * pulExtendedErrorCode )
+{
+    AzureIoTProvisioningClientResult_t xReturn;
+
+    if( ( xProvClientHandle == NULL ) ||
+        ( pulExtendedErrorCode == NULL ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning client extended code get failed : Invalid argument" ) );
```
ericwol-msft, line 949:
```suggestion
        AZLogError( ( "AzureIoTProvisioning client get extended code failed: Invalid argument" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+AzureIoTProvisioningClientResult_t AzureIoTProvisioningClient_SetRegistrationPayload( AzureIoTProvisioningClientHandle_t xProvClientHandle,
+                                                                                      const uint8_t * pucPayload,
+                                                                                      uint32_t ulPayloadLength )
+{
+    AzureIoTProvisioningClientResult_t xReturn;
+
+    if( ( xProvClientHandle == NULL ) ||
+        ( pucPayload == NULL ) || ( ulPayloadLength == 0 ) )
+    {
+        AZLogError( ( "AzureIoTProvisioning client registration payload failed : Invalid argument" ) );
```
ericwol-msft, line 977:
```suggestion
        AZLogError( ( "AzureIoTProvisioning setting client registration payload failed: Invalid argument" ) );
```

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,994 @@
...
+        xReturn = eAzureIoTProvisioningInvalidArgument;
+    }
+    else if( xProvClientHandle->_internal.ulWorkflowState != azureiotprovisioningWF_STATE_INIT )
+    {
+        AZLogError( ( "AzureIoTProvisioning client state is not in init" ) );
```
ericwol-msft, line 982:
```suggestion
        AZLogError( ( "AzureIoTProvisioning client state is not initialized" ) );
```

ericwol-msft (approved):

source/azure_iot_provisioning_client.c
hihigupt: INIT state required for set to work properly

---
In reply to: [603655561](https://github.com/Azure/azure-iot-middleware-freertos/pull/49#discussion_r603655561) [](ancestors = 603655561)

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
jspaith: "I will remove the \r\n from logs and let the log implementation to add new line."
Yeah, I think that's fair and if someone had a platform it mattered they could always tweak there.  It's a very minor perf hit on 2 I/O calls presumably but it's not that bad and logging doesn't promise ultra high perf anyway.

I took a quick look through FreeRTOS source and they're all over the place about consistency - I even saw a "\n\r" which in +20 years of doing I've never seen before:).

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
jspaith: OK - to keep simple agreed.  I'll resolve this conv then.

source/azure_iot_provisioning_client.c
jspaith: Correct, I see the `azureiotprovisioningWF_STATE_COMPLETE` comparison.  My question is whether that's the correct check?  Maybe we should allow this for both azureiotprovisioningWF_STATE_COMPLETE and for `azureiotprovisioningWF_STATE_WAITING` also for scenario that service said "300021 code, waitfor=2 minutes".  So app may care about the 300021 status and to determine if they want to go through the wait require loop or not.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,992 @@
...
+        }
+
+        xReturn = prvProvClientRunWorkflow( xProvClientHandle, ulTimeoutMilliseconds );
+    }
+
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTProvisioningClientResult_t AzureIoTProvisioningClient_GetHub( AzureIoTProvisioningClientHandle_t xProvClientHandle,
```
jspaith, line 886: Should we call this AzureIoTProvisioningClient_GetDeviceAndHub, since it's returning both?

jspaith (approved):

source/azure_iot_provisioning_client.c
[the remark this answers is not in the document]
hihigupt: Closing this for now

source/azure_iot_provisioning_client.c
hihigupt: Currently I have added in doc: only call this once registration is done. Once done, this API will return last error code retrun from Service.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,993 @@
...
+        ulDeviceIDLength = ( uint32_t ) az_span_size( *pxDeviceID );
+
+        if( ( *pulHostnameLength < ulHostnameLength ) || ( *pulDeviceIDLength < ulDeviceIDLength ) )
+        {
+            AZLogWarn( ( "AzureIoTProvisioning Memory buffer passed is not enough to store hub info" ) );
+            xReturn = eAzureIoTProvisioningFailed;
+        }
+        else
+        {
+            memcpy( pucHubHostname, az_span_ptr( *pxHostname ), ulHostnameLength );
```
CIPop, line 928: Same question about memory copying.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,992 @@
...
+    az_json_writer xJsonWriter;
+    az_span xBuffer = az_span_create( pucPayload, ( int32_t ) ulPayloadLength );
+    az_span xRegistrationIdLabel = AZ_SPAN_LITERAL_FROM_STR( azureiotprovisioningREQUEST_REGISTRATION_ID_LABEL );
+    az_span xRegistrationId = az_span_create( ( uint8_t * ) xProvClientHandle->_internal.pucRegistrationID,
+                                              ( int32_t ) xProvClientHandle->_internal.ulRegistrationIDLength );
+    az_span xPayloadLabel = AZ_SPAN_LITERAL_FROM_STR( azureiotprovisioningREQUEST_PAYLOAD_LABEL );
+    az_span xPayload = az_span_create( ( uint8_t * ) xProvClientHandle->_internal.pucRegistrationPayload,
+                                       ( int32_t ) xProvClientHandle->_internal.ulRegistrationPayloadLength );
+
+    if( az_result_failed( az_json_writer_init( &xJsonWriter, xBuffer, NULL ) ) ||
```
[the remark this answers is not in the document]
hihigupt: I already told @danewalton that this does not belong here, and I think he will work on this to surface it from embedded sdk. Once, that is there we can remove this.

source/azure_iot_provisioning_client.c
```diff
@@ -0,0 +1,992 @@
...
+    if( xProvClientHandle->_internal.ulWorkflowState == azureiotprovisioningWF_STATE_REQUESTING )
+    {
+        if( ( pxPublishInfo->xPayloadLength + pxPublishInfo->usTopicNameLength ) <=
+            sizeof( xProvClientHandle->_internal.ucProvisioningLastResponse ) )
+        {
+            /* Copy topic + payload into the response buffer.
+             * ucProvisioningLastResponse = [ topic payload ] */
+            xProvClientHandle->_internal.usLastResponseTopicLength =
+                pxPublishInfo->usTopicNameLength;
+            memcpy( xProvClientHandle->_internal.ucProvisioningLastResponse,
```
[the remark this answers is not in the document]
hihigupt: MQTT stack owns the memory and span depends does not take the ownership. So that is the reason middleware copy to memory that it owns and access it outside mqtt callback.

source/azure_iot_provisioning_client.c
hihigupt: Same.

PR #50, opened 2021-03-31
Add default config

hihigupt:


source/include/azure_iot_config_defaults.h
```diff
@@ -0,0 +1,124 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+#ifndef AZURE_IOT_CONFIG_DEFAULT_H
+#define AZURE_IOT_CONFIG_DEFAULT_H
```
ericwol-msft, line 5:
```suggestion
#define AZURE_IOT_CONFIG_DEFAULTS_H
```

source/include/azure_iot_config_defaults.h
```diff
@@ -0,0 +1,124 @@
...
+ * <b>Default value</b>: Debug logging is turned off, and no code is generated for calls
+ * to the macro in the AzureIoT Middleware library on compilation.
+ */
+#ifndef LogDebug
+    #define AZLogDebug( message )
+#else
+    #define AZLogDebug       LogDebug
+#endif
+
+#endif /* AZURE_IOT_CONFIG_DEFAULT_H */
```
ericwol-msft, line 124:
```suggestion
#endif /* AZURE_IOT_CONFIG_DEFAULTS_H */
```

ericwol-msft (approved):

source/include/azure_iot_config_defaults.h
```diff
@@ -0,0 +1,124 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+#ifndef AZURE_IOT_CONFIG_DEFAULTS_H
+#define AZURE_IOT_CONFIG_DEFAULTS_H
+
+#ifndef azureiotDEFAULT_TOKEN_TIMEOUT_IN_SEC
```
jspaith, line 7: Even though these are somewhat self-documenting, I feel some additional doxygen would help here so there's not ambiguity.

source/include/azure_iot_config_defaults.h
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * @brief Milliseconds per second.
+ */
+#define azureiotMILLISECONDS_PER_SECOND                 ( 1000U )
+
+/**
+ * @brief Macro that is called in the AzureIoT Middleware library for logging "Error" level
+ * messages.
+ *
+ * To enable error level logging in the AzureIoT Middleware library, this macro should be mapped to the
```
jspaith, line 52: It's a bit more nuanced, because I can also enable enable this by `#define LogError` where I'm not necessarily the one mapping (which I'd think as my app code doing the `#define AZLogError <whatever>`).

Maybe something like below, with your word-smithing love of course and making sure my understanding of the FreeRTOS log setup is right.  I don't think our docs should explicitly say here #define LIBRARY_LOG_LEVEL or above, because that might change (albeit unlikely) but we can at least get folks in right direction.
```
To enable error level logging in the AzureIoT Middleware library either:
* Map the macro to the application-specific logging implementation that supports error logging.
* Enable FreeRTOS's logging stack by `#include <logging_stack.h>` and configuring its logging level to be at least error level.
``

Similar feedback for other log levels.

File: source/include/azure_iot_config_defaults.h [hunk 5]
```diff
@@ -0,0 +1,124 @@
...
+ * application-specific logging implementation that supports error logging.
+ *
+ * @note This logging macro is called in the AzureIoT Middleware library with parameters wrapped in
+ * double parentheses to be ISO C89/C90 standard compliant.
+ *
+ * <b>Default value</b>: Error logging is turned off, and no code is generated for calls
+ * to the macro in the AzureIoT Middleware library on compilation.
```
jspaith, line 59: "AzureIoT" => "Azure IoT" here & other places.

jspaith (approved):
Some nits but generally looks good.

PR #58, opened 2021-04-02
azure_iot.c file

danewalton:


source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+static AzureIoTResult_t prvAzureIoTBase64Decode( char * base64name,
+                                                uint32_t length,
+                                                uint8_t * name,
+                                                uint32_t name_size,
+                                                uint32_t * bytes_copied )
+{
+    uint32_t i, j;
+    uint32_t value1, value2;
+    uint32_t step;
+    uint32_t sourceLength = length;
```
hihigupt, line 36: rename all based code conventions

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+
+    /* Put a NULL character in.  */
+    name[ j ] = 0;
+    *bytes_copied = j;
+
+    return eAzureIoTSuccess;
+}
+/*-----------------------------------------------------------*/
+
+static AzureIoTResult_t prvAzureIoTBase64Encode( uint8_t * name,
```
hihigupt, line 150: all private function, can we have comment at the start of it.

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/**
+ * @file azure_iot.c
+ * @brief -------.
```
hihigupt, line 6: add some description

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+{
+    ( void ) classification;
+    //In case logs are stripped out, suppress unused parameter error.
+    ( void ) message;
+
+    AZLogInfo( ( "%.*s", az_span_size( message ), az_span_ptr( message ) ) );
+}
+/*-----------------------------------------------------------*/
+
+static AzureIoTResult_t prvAzureIoTBase64Decode( char * base64name,
```
hihigupt, line 27: add todo: to replace it with sdk

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+        if( step == 0 )
+        {
+            /* Use first 6 bits of name character for index.  */
+            base64name[ j++ ] = ( char ) _azure_iot_base64_array[ ( ( uint8_t ) name[ i ] ) >> 2 ];
+            step++;
+        }
+        else if( step == 1 )
+        {
+            /* Use last 2 bits of name character and first 4 bits of next name character for index.  */
+            base64name[ j++ ] = ( char ) _azure_iot_base64_array[ ( ( ( ( uint8_t ) name[ i ] ) & 0x3 ) << 4 ) | ( ( ( uint8_t ) name[ i + 1 ] ) >> 4 ) ];
```
hihigupt, line 204: lets try to limit character of line by 120

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+                                              uint32_t ulMessageSize,
+                                              uint8_t * pucBuffer,
+                                              uint32_t ulBufferLength,
+                                              uint8_t * pucOutput,
+                                              uint32_t ulOutputSize,
+                                              uint32_t * pulOutputLength )
+{
+    AzureIoTResult_t status;
+    uint8_t * hash_buf;
+    uint32_t hash_buf_size = 33;
```
hihigupt, line 359: lets make macro for magic numbers

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+}
+/*-----------------------------------------------------------*/
+
+static AzureIoTResult_t prvAzureIoTBase64Decode( char * pcBase64name,
+                                                uint32_t ulLength,
+                                                uint8_t * pucName,
+                                                uint32_t ulNameSize,
+                                                uint32_t * pulBytesCopied )
+{
+    uint32_t i, j;
```
hihigupt, line 33: lets name these variable too with the coding style.

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/**
+ * @file azure_iot.c
+ * @brief -------.
+ *
+ */
+
```
hihigupt, line 9: should we include header for memset ?

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+static const char _cAzureIoTBase64Array[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+/*-----------------------------------------------------------*/
+
+static void prvAzureIoTLogListener( az_log_classification xClassification,
+                                    az_span xMessage )
+{
+    ( void ) xClassification;
+    //In case logs are stripped out, suppress unused parameter error.
```
hihigupt, line 20: comments should use /* */

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+
+static AzureIoTResult_t prvAzureIoTBase64Encode( uint8_t * pucName,
+                                                uint32_t ulLength,
+                                                char * pcBase64name,
+                                                uint32_t base64name_size )
+{
+    uint32_t ulPad;
+    uint32_t i, j;
+    uint32_t ulStep;
+
```
hihigupt, line 158: remove extra line

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+                                                uint8_t * pucBuffer,
+                                                uint32_t ulAlreadyWrittenLength,
+                                                uint32_t ulBufferLength )
+{
+    az_span xPropertyBufferSpan = az_span_create( pucBuffer, ( int32_t ) ulBufferLength );
+    az_result xResult;
+
+    if( pxMessageProperties == NULL )
+    {
+        AZLogError( ( "AzureIoTMessagePropertiesInit failed: Invalid argument \r\n" ) );
```
hihigupt, line 270: I remember John pointing out that we should be consistent with our logging to be ending with \r\n or not... in provisioning I removed all the \r\n. 
I am ok with either as long we are consistent with our story.

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+/*-----------------------------------------------------------*/
+
+AzureIoTResult_t AzureIoT_MessagePropertiesFind( AzureIoTMessageProperties_t * pxMessageProperties,
+                                                const uint8_t * pucName,
+                                                uint32_t ulNameLength,
+                                                const uint8_t ** ppucOutValue,
+                                                uint32_t * pulOutValueLength )
+{
+    az_span xNameSpan = az_span_create( (uint8_t * ) pucName, ( int32_t ) ulNameLength );
+    az_span outValueSpan;
```
hihigupt, line 321: xOutValueSpan

source/azure_iot.c
```diff
@@ -0,0 +1,410 @@
...
+
+    xStatus = prvAzureIoTBase64Encode( pucHashBuf, ulHashBufSize,
+                                      ( char * ) pucOutput, ulOutputSize );
+
+    if( xStatus )
+    {
+        return xStatus;
+    }
+
+    *pulOutputLength = ( uint32_t ) strlen( ( char * )pucOutput );
```
hihigupt, line 406: can we return length from prvAzureIoTBase64Encode so that we do not have to link strlen?

source/azure_iot.c
danewalton: removing is a good idea

source/azure_iot.c
danewalton: ehhh do we have to here? I understand it's not consistent but I feel like this is a pretty common practice. Otherwise its `ulI` and `ulJ` or something like that.

source/azure_iot.c
danewalton: yea good call

source/azure_iot.c
```diff
@@ -0,0 +1,438 @@
...
+
+    /* Now add the PAD characters.  */
+    while( ( ulPad-- ) && ( j < base64name_size ) )
+    {
+        /* Pad pcBase64name with '=' characters.  */
+        pcBase64name[ j++ ] = '=';
+    }
+
+    /* Put a NULL character in.  */
+    pcBase64name[ j ] = 0;
```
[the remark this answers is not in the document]
jspaith: This is what we discussed today on call.

source/azure_iot.c
```diff
@@ -0,0 +1,438 @@
...
+    /* Compute the pcBase64name.  */
+    ulStep = 0;
+    i = 0;
+
+    while( j < ulLength )
+    {
+        /* Determine which step we are in.  */
+        if( ulStep == 0 )
+        {
+            /* Use first 6 bits of name character for index.  */
```
jspaith, line 221: What happens if I pass in the string `char* testString = "\xFFThis is a test";`?  From my math, the \0xFF when bit-shifted >>2 will turn into 63.  But the problem is the array we're dereferencing is only 33 so we'll go off the end.

I don't see any guidance about what a legal string is, never mind checking.

source/azure_iot.c
```diff
@@ -0,0 +1,438 @@
...
+    ulBinaryKeyBufSize = ulBufferLength;
+    xStatus = prvAzureIoTBase64Decode( ( char * ) pucKey, ulKeySize,
+                                       pucBuffer, ulBinaryKeyBufSize, &ulBinaryKeyBufSize );
+
+    if( xStatus )
+    {
+        return xStatus;
+    }
+
+    ulBufferLength -= ulBinaryKeyBufSize;
```
jspaith, line 408: Nit: Whenever there's subtraction I always worry we'll end up with underflow, so ulBufferLength would be negative (or giant unsigned int) if ulBinaryKeyBufSize is bigger.

That's not case here because the decoded length will be smaller, but it might not be bad to assert if we do such things about this or put a comment to this effect?

source/azure_iot.c
```diff
@@ -0,0 +1,438 @@
...
+                                               const uint8_t * pucMessage,
+                                               uint32_t ulMessageSize,
+                                               uint8_t * pucBuffer,
+                                               uint32_t ulBufferLength,
+                                               uint8_t * pucOutput,
+                                               uint32_t ulOutputSize,
+                                               uint32_t * pulOutputLength )
+{
+    AzureIoTResult_t xStatus;
+    uint8_t * pucHashBuf;
```
jspaith, line 385: Nit - consider a new variable for `uint8_t * pucDecodedKeyBuf = pucBuffer;` and then use `pucDecodedKeyBuf` instead of `pucBuffer` when calculating and using the key.

One the one hand I hesitate on this because it's aliasing.  But on other hand you have a similar pattern with `pucHashBuf` pointing to the place in the passed user buffer where it needs to write to.  

I suggest above because the intent is much clearer with the name to not just be buffer.

source/azure_iot.c
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    *ppucOutValue = az_span_ptr( outValueSpan );
+    *pulOutValueLength = ( uint32_t ) az_span_size( outValueSpan );
+
+    return eAzureIoTSuccess;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTResult_t AzureIoT_Base64HMACCalculate( AzureIoTGetHMACFunc_t xAzureIoTHMACFunction,
```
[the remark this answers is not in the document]
danewalton: tracking https://github.com/Azure/azure-iot-middleware-freertos/issues/64

source/azure_iot.c
```diff
@@ -0,0 +1,412 @@
...
+}
+/*-----------------------------------------------------------*/
+
+void AzureIoT_Deinit()
+{
+    AZLogDebug( ( "AzureIoT_Deinit called \r\n" ) );
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTResult_t AzureIoT_MessagePropertiesInit( AzureIoTMessageProperties_t * pxMessageProperties,
```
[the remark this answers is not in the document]
danewalton: good point. we should probably discuss here. @hihigupt you have any opinion? I don't see any reason why we can't just make it a pointer only (no handle) since we don't do any malloc'ing?

source/azure_iot.c
```diff
@@ -0,0 +1,412 @@
...
+                                                 uint8_t * pucBuffer,
+                                                 uint32_t ulAlreadyWrittenLength,
+                                                 uint32_t ulBufferLength )
+{
+    az_span xPropertyBufferSpan = az_span_create( pucBuffer, ( int32_t ) ulBufferLength );
+    az_result xResult;
+
+    if( pxMessageProperties == NULL )
+    {
+        AZLogError( ( "AzureIoTMessagePropertiesInit failed: Invalid argument \r\n" ) );
```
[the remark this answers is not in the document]
danewalton: will check them all

source/azure_iot.c
```diff
@@ -0,0 +1,438 @@
...
+
+/**
+ *
+ * Decode Base64 bytes.
+ *
+ * TODO: Remove in favor of embedded SDK implementation.
+ *
+ * */
+static AzureIoTResult_t prvAzureIoTBase64Decode( char * pcBase64name,
+                                                 uint32_t ulLength,
```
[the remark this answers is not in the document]
danewalton: updated all the var names

source/azure_iot.c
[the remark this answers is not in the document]
hihigupt: Making an extra type for pointer to control block is fine with me.

.gitignore
```diff
@@ -0,0 +1,64 @@
+# Prerequisites
+*.d
+
+# Object files
+*.o
```
hihigupt, line 5: this might not be required now

source/azure_iot.c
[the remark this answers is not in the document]
danewalton: discussed offline. Removing `_Handle` types since user allocates the structs and hands them to us. Giving us pointers to structs makes sense in that case. If we were `malloc`'ing and returning to the user it would probably make more sense to use handles.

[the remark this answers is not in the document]
danewalton: https://github.com/Azure/azure-iot-middleware-freertos/pull/67
PR to tackle

source/azure_iot.c
danewalton: resolving with not changing unless someone is really opposed

source/azure_iot.c
```diff
@@ -0,0 +1,438 @@
...
+                                               uint8_t * pucOutput,
+                                               uint32_t ulOutputSize,
+                                               uint32_t * pulOutputLength )
+{
+    AzureIoTResult_t xStatus;
+    uint8_t * pucHashBuf;
+    uint32_t ulHashBufSize = AZURE_BASE64_HASH_BUFFER_SIZE;
+    uint32_t ulBinaryKeyBufSize;
+
+    if( ( xAzureIoTHMACFunction == NULL ) ||
```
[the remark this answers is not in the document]
danewalton: yea for now since this is public we check. the others are static so no go on those. These will be properly updated if we change the visibility of the function

source/azure_iot.c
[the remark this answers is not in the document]
danewalton: Addressed here let me know what you think:
https://github.com/Azure/azure-iot-middleware-freertos/pull/58/commits/1066fc435318a02568396a588a193422e4e5c190

source/azure_iot.c
danewalton: Just double checking here. The array we index into is `64` long. Can you clarify here?

source/azure_iot.c
danewalton: fair. makes readability easier

.gitignore
danewalton: yea I think i rebased and fixed

source/azure_iot.c
danewalton: added a comment. let me know if it's sufficient

source/azure_iot.c
```diff
@@ -0,0 +1,455 @@
...
+
+/**
+ * @file azure_iot.c
+ *
+ * @brief Common Azure IoT code for FreeRTOS middleware.
+ *
+ */
+
+#include <stdio.h>
+
```
hihigupt, line 12: still required?

source/azure_iot.c
```diff
@@ -0,0 +1,455 @@
...
+#include "azure_iot.h"
+
+#define azureiotBASE64_HASH_BUFFER_SIZE    33
+
```
hihigupt, line 16: please use ( 33 )

source/azure_iot.c
```diff
@@ -0,0 +1,455 @@
...
+    #endif
+
+    return eAzureIoTSuccess;
+}
+/*-----------------------------------------------------------*/
+
+void AzureIoT_Deinit()
+{
+    AZLogDebug( ( "AzureIoT_Deinit called" ) );
+}
```
hihigupt, line 300: Not useful debug statement, can we remove this.

source/azure_iot.c
jspaith: Oops.  Yup, I misread the (33) above and thought it was array len.  This is totally cool as-is.  Sorry about that.

source/azure_iot.c
jspaith: yup, resolving.

source/azure_iot.c
```diff
@@ -0,0 +1,455 @@
...
+                                                 uint32_t ulBufferLength )
+{
+    az_span xPropertyBufferSpan = az_span_create( pucBuffer, ( int32_t ) ulBufferLength );
+    az_result xResult;
+
+    if( pxMessageProperties == NULL )
+    {
+        AZLogError( ( "AzureIoT_MessagePropertiesInit failed: Invalid argument" ) );
+        return eAzureIoTInvalidArgument;
+    }
```
hihigupt, line 315: Can we try to have one return from function.  Atleast in simple functions

jspaith (approved):

source/azure_iot.c
```diff
@@ -0,0 +1,455 @@
...
+
+    if( xAzureIoTHMACFunction( pucDecodedKeyBuf, ulBinaryKeyBufSize,
+                               pucMessage, ( uint32_t ) ulMessageSize,
+                               pucHashBuf, ulHashBufSize, &ulHashBufSize ) )
+    {
+        return eAzureIoTFailed;
+    }
+
+    uint32_t ulBase64OutputLength;
+    xStatus = prvAzureIoTBase64Encode( pucHashBuf, ulHashBufSize,
```
hihigupt, line 443: lets define all variable at start of function. It really help in code review and debugging to know how much stack, function is using. (more closer to what compiler will do). Similar to AZRTOS

hihigupt: and FreeRTOS too.

---
In reply to: [608244686](https://github.com/Azure/azure-iot-middleware-freertos/pull/58#discussion_r608244686) [](ancestors = 608244686)

source/azure_iot.c
hihigupt: Can we add comment that we are using SHA256 so we need 32 bytes

---
In reply to: [608240846](https://github.com/Azure/azure-iot-middleware-freertos/pull/58#discussion_r608240846) [](ancestors = 608240846)

source/azure_iot.c
danewalton: hmmm yea I don't know what this is here tbh

danewalton: oh wait it's memset

danewalton: replacing with `string.h`

source/azure_iot.c
danewalton: yes and yes

source/azure_iot.c
danewalton: yea usually I would say this is a religious debate but if it's similar to the others then I'm good with it

source/azure_iot.c
hihigupt: just to make sure that we are not causing any warning ... Please add this file to test cmake

https://github.com/Azure/azure-iot-middleware-freertos/blob/332117bbcbd4ef1592b3c275652d5c1e4335681f/tests/CMakeLists.txt#L28

source/azure_iot.c
danewalton: I understand the sentiment. Only problem I have is that gets to be wishy-washy with PR's. There will be debates about what constitutes "simple enough" for one return. Ideally I think we should either accept multiple returns or just have one. From a future maintainability standpoint.

hihigupt (approved):

PR #60, opened 2021-04-02
add hub client c file

danewalton:
I backported all the changes from the header review into this file. I have it compiling in another branch here:
https://github.com/Azure/azure-iot-middleware-freertos/tree/dawalton/backporthubclient


source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+/*
+ * Topic subscribe state
+ */
+#define azureiothubTOPIC_SUBSCRIBE_STATE_NONE       0
+#define azureiothubTOPIC_SUBSCRIBE_STATE_SUB        1
+#define azureiothubTOPIC_SUBSCRIBE_STATE_SUBACK     2
+
+/*-----------------------------------------------------------*/
+
+static void prvMQTTProcessIncomingPublish( AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
```
danewalton, line 44: add comments for all `prv` functions

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+        }
+
+        if( az_result_failed( az_iot_hub_client_init( &xAzureIoTHubClientHandle->_internal.xAzureIoTHubClientCore,
+                                                      hostname_span, device_id_span, &options ) ) )
+        {
+            AZLogError( ( "Failed to initialize az_iot_hub_client_init \r\n" ) );
+            ret = eAzureIoTHubClientInitFailed;
+        }
+        /* Initialize AzureIoTMQTT library. */
+        else if( ( AzureIoTMQTT_Init( &( xAzureIoTHubClientHandle->_internal.xMQTTContext ), pxTransportInterface,
```
jspaith, line 353: We should store AzureIoTMQTTResult_t from this call and on failure output value to log.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+    }
+    else
+    {
+        memset( ( void * ) xAzureIoTHubClientHandle, 0, sizeof( AzureIoTHubClient_t ) );
+
+        /* Setup scratch buffer to be used by middleware */
+        xAzureIoTHubClientHandle->_internal.ulAzureIoTHubClientWorkingBufferLength =
+            ( azureiotUSERNAME_MAX + azureiotPASSWORD_MAX ) > azureiotTOPIC_MAX ? ( azureiotUSERNAME_MAX + azureiotPASSWORD_MAX ) : azureiotTOPIC_MAX;
+        xAzureIoTHubClientHandle->_internal.pucAzureIoTHubClientWorkingBuffer = pucBuffer;
+        pucBuffer += xAzureIoTHubClientHandle->_internal.ulAzureIoTHubClientWorkingBufferLength;
```
jspaith, line 326: Could we create a new variable for `pucBuffer` so intent later on is more clear?  I'm making up (bad) name but like below.

This keeps pucBuffer which is passed the same, and any reasonable compiler optimizer won't have trouble re-using register for `pucBuffer` and a new variable `pucNetworkBuffer`.

```c
uint8_t * pucNetworkBuffer = pucBuffer + xAzureIoTHubClientHandle->_internal.ulAzureIoTHubClientWorkingBufferLength;
// Then pass the pucNetworkBuffer to AzureIoTMQTT_Init
```

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+/*-----------------------------------------------------------*/
+
+AzureIoTHubClientResult_t AzureIoTHubClient_Connect( AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
+                                                     bool cleanSession,
+                                                     uint32_t ulTimeoutMilliseconds )
+{
+    AzureIoTMQTTConnectInfo_t xConnectInfo = { 0 };
+    AzureIoTHubClientResult_t ret;
+    AzureIoTMQTTStatus_t xResult;
+    bool xSessionPresent;
```
jspaith, line 388: I think we need to plumb this up to the caller, don't we?  If the session isn't present they're going to need to behave differently (re-subscribe) than if session is there and they asked it to be.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+        else
+        {
+            xConnectInfo.cleanSession = cleanSession;
+            xConnectInfo.pClientIdentifier = ( const char * ) xAzureIoTHubClientHandle->_internal.pucDeviceID;
+            xConnectInfo.clientIdentifierLength = ( uint16_t ) xAzureIoTHubClientHandle->_internal.ulDeviceIDLength;
+            xConnectInfo.userNameLength = ( uint16_t ) mqtt_user_name_length;
+            xConnectInfo.keepAliveSeconds = azureiothubKEEP_ALIVE_TIMEOUT_SECONDS;
+            xConnectInfo.passwordLength = ( uint16_t ) bytesCopied;
+
+            /* Send MQTT CONNECT packet to broker. LWT is not used in this demo, so it
```
jspaith, line 432: 1 - If we need this, please expand out LWT to Last Will and Testament.  I get it should be obvious but there's so many acronyms out there...
2 - This isn't a demo anymore!

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+
+AzureIoTHubClientResult_t AzureIoTHubClient_Connect( AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
+                                                     bool cleanSession,
+                                                     uint32_t ulTimeoutMilliseconds )
+{
+    AzureIoTMQTTConnectInfo_t xConnectInfo = { 0 };
+    AzureIoTHubClientResult_t ret;
+    AzureIoTMQTTStatus_t xResult;
+    bool xSessionPresent;
+    uint32_t bytesCopied = 0;
```
jspaith, line 389: Recommend variable name `bytesCopied` reflecting ultimate intent, like `passwordLength`.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+        xMQTTPublishInfo.qos = AzureIoTMQTTQoS1;
+        xMQTTPublishInfo.pTopicName = ( const char * ) xAzureIoTHubClientHandle->_internal.pucAzureIoTHubClientWorkingBuffer;
+        xMQTTPublishInfo.topicNameLength = ( uint16_t ) telemetry_topic_length;
+        xMQTTPublishInfo.pPayload = ( const void * ) pucTelemetryData;
+        xMQTTPublishInfo.payloadLength = ulTelemetryDataLength;
+
+        /* Get a unique packet id. */
+        usPublishPacketIdentifier = AzureIoTMQTT_GetPacketId( &( xAzureIoTHubClientHandle->_internal.xMQTTContext ) );
+
+        /* Send PUBLISH packet. Packet ID is not used for a QoS1 publish. */
```
jspaith, line 524: I don't understand comment about PackeID not used, but we use it on the next line.  I thought it was needed for PUBACK?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+/*-----------------------------------------------------------*/
+
+static void prvMQTTProcessResponse( AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
+                                    AzureIoTMQTTPacketInfo_t * pxIncomingPacket,
+                                    uint16_t usPacketId )
+{
+    uint32_t index;
+    AzureIoTHubClientReceiveContext_t * context;
+
+    if( ( pxIncomingPacket->type & 0xF0U ) == AZURE_IOT_MQTT_PACKET_TYPE_SUBACK )
```
jspaith, line 74: How do we process PUBACK's and pass up to application?  I get it may not yet be plumbed yet, but even looking at the public header of this it's not clear how this is supposed ot even work in 1st place

From my memory and a very quick review of CoreMQTT, I don't think the publish call blocks until the PUBACK is received but just till bytes get sent across.  It seems -again quick - in CoreMQTT than in `handlePublishAcks` is where this is plumbed up to app but that `handlePublishAcks` is itself in the receive loop someplace.

jspaith (changes requested):

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+    {
+        for( index = 0; index < ( sizeof( xAzureIoTHubClientHandle->_internal.xReceiveContext ) / sizeof( xAzureIoTHubClientHandle->_internal.xReceiveContext[ 0 ] ) ); index++ )
+        {
+            context = &xAzureIoTHubClientHandle->_internal.xReceiveContext[ index ];
+
+            if( context->_internal.usMqttSubPacketID == usPacketId )
+            {
+                /* TODO: inspect packet to see is ack was successful*/
```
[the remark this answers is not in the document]
danewalton: Gotcha I'll try and remember in the future to do it too.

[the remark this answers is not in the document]
danewalton: https://github.com/Azure/azure-iot-middleware-freertos/issues/65

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+
+#ifndef azureiothubUSER_AGENT
+    #define azureiothubUSER_AGENT    "DeviceClientType=c%2F" AZ_SDK_VERSION_STRING "%28FreeRTOS%29"
+#endif /* azureiothubUSER_AGENT */
+
+/*
+ * Topic subscribe state
+ */
+#define azureiothubTOPIC_SUBSCRIBE_STATE_NONE       0
+#define azureiothubTOPIC_SUBSCRIBE_STATE_SUB        1
```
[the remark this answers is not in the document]
danewalton: yea i tried removing everything that isn't relevant but could have missed some stuff. I think when I did a "find all" in the doc I quickly mistook a hit on `_SUB` with `_SUBACK`

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+        prvMQTTProcessIncomingPublish( xAzureIoTHubClientHandle, pxDeserializedInfo->pPublishInfo );
+    }
+    else
+    {
+        prvMQTTProcessResponse( xAzureIoTHubClientHandle, pxPacketInfo, pxDeserializedInfo->packetIdentifier );
+    }
+}
+/*-----------------------------------------------------------*/
+
+static uint32_t prvAzureIoTHubClientCloudMessageProcess( AzureIoTHubClientReceiveContext_t * context,
```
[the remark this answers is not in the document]
danewalton: yea I agree. I think with private functions I'm inclined to make it `c2d` for shorthand

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+                                                         AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
+                                                         void * pxPublishInfo )
+{
+    AzureIoTHubClientCloudToDeviceMessageRequest_t message;
+    AzureIoTMQTTPublishInfo_t * mqttPublishInfo = ( AzureIoTMQTTPublishInfo_t * ) pxPublishInfo;
```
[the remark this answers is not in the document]
danewalton: yea definitely need to go through and update the var names. unfortunately preliminarily it was just in the API's and now needs to be propagated. will do

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+
+static void prvMQTTProcessIncomingPublish( AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
+                                           AzureIoTMQTTPublishInfo_t * pxPublishInfo )
+{
+    uint32_t index;
+
+    configASSERT( pxPublishInfo != NULL );
+    AzureIoTHubClientReceiveContext_t * context;
+
+    for( index = 0; index < ( sizeof( xAzureIoTHubClientHandle->_internal.xReceiveContext ) / sizeof( xAzureIoTHubClientHandle->_internal.xReceiveContext[ 0 ] ) ); index++ )
```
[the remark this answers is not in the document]
danewalton: updated the name and changed to use the macro good callout

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+#ifndef azureiothubKEEP_ALIVE_TIMEOUT_SECONDS
+    #define azureiothubKEEP_ALIVE_TIMEOUT_SECONDS    azureiotKEEP_ALIVE_TIMEOUT_SECONDS
+#endif /* azureiothubKEEP_ALIVE_TIMEOUT_SECONDS */
+
+#ifndef azureiothubSUBACK_WAIT_INTERVAL_MS
+    #define azureiothubSUBACK_WAIT_INTERVAL_MS    azureiotSUBACK_WAIT_INTERVAL_MS
+#endif /* azureiothubSUBACK_WAIT_INTERVAL_MS */
+
+#ifndef azureiothubUSER_AGENT
+    #define azureiothubUSER_AGENT    "DeviceClientType=c%2F" AZ_SDK_VERSION_STRING "%28FreeRTOS%29"
```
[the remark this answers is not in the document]
danewalton: https://github.com/Azure/azure-iot-middleware-freertos/issues/72
Tracking here. I agree but that update will make this larger so I'll leave to tackle once the impl is merged.

source/azure_iot_hub_client.c
danewalton: Yea I think this is an erroneous comment. @hihigupt to confirm.

https://github.com/FreeRTOS/coreMQTT/blob/f756d9833ad79b60084def59935fdabf475c08b0/source/include/core_mqtt.h#L501

source/azure_iot_hub_client.c
danewalton: agreed. changing

source/azure_iot_hub_client.c
hihigupt: Wrong comment

---
In reply to: [608951348](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r608951348) [](ancestors = 608951348)

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+                              AzureIoTMQTTDeserializedInfo_t * pxDeserializedInfo )
+{
+    /* First element in AzureIoTHubClientHandle */
+    AzureIoTHubClientHandle_t xAzureIoTHubClientHandle = ( AzureIoTHubClientHandle_t ) pxMQTTContext;
+
+    if( ( pxPacketInfo->type & 0xF0U ) == AZURE_IOT_MQTT_PACKET_TYPE_PUBLISH )
+    {
+        prvMQTTProcessIncomingPublish( xAzureIoTHubClientHandle, pxDeserializedInfo->pPublishInfo );
+    }
+    else
```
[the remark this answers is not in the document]
danewalton: updated

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+static uint32_t prvAzureIoTHubClientCloudMessageProcess( AzureIoTHubClientReceiveContext_t * context,
+                                                         AzureIoTHubClientHandle_t xAzureIoTHubClientHandle,
+                                                         void * pxPublishInfo )
+{
+    AzureIoTHubClientCloudToDeviceMessageRequest_t message;
+    AzureIoTMQTTPublishInfo_t * mqttPublishInfo = ( AzureIoTMQTTPublishInfo_t * ) pxPublishInfo;
+    az_iot_hub_client_c2d_request out_request;
+    az_span topic_span = az_span_create( ( uint8_t * ) mqttPublishInfo->pTopicName, mqttPublishInfo->topicNameLength );
+
+    if( az_result_failed( az_iot_hub_client_c2d_parse_received_topic( &xAzureIoTHubClientHandle->_internal.xAzureIoTHubClientCore,
```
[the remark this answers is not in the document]
danewalton: that sounds good for me

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+                                                                      topic_span, &out_request ) ) )
+    {
+        return eAzureIoTHubClientFailed;
+    }
+
+    /* Process incoming Publish. */
+    AZLogInfo( ( "Cloud message QoS : %d\n", mqttPublishInfo->qos ) );
+
+    /* Verify the received publish is for the we have subscribed to. */
+    AZLogInfo( ( "Cloud message topic: %.*s  with payload : %.*s",
```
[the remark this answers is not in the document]
hihigupt: how about moving this to Debug level?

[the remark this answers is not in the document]
danewalton: https://github.com/Azure/azure-iot-middleware-freertos/issues/74
PR to track/discuss

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/**
+ * @file azure_iot_hub_client.c
+ * @brief -------.
+ */
```
hihigupt, line 7: brief?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
...
+
+#include "azure_iot_hub_client.h"
+#include "azure_iot_mqtt.h"
```
hihigupt, line 10: >azure_iot_mqtt [](start = 10, length = 14)

move it to line 16?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
...
+
+/*-----------------------------------------------------------*/
+
+static void prvMQTTProcessIncomingPublish( AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                           AzureIoTMQTTPublishInfo_t * pxPublishInfo )
+{
+    uint32_t ulIndex;
+
+    configASSERT( pxPublishInfo != NULL );
+    AzureIoTHubClientReceiveContext_t * pxContext;
```
hihigupt, line 50: move it to line 48

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
...
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ ulIndex ];
+
+        if( ( pxContext->_internal.pxProcessFunction != NULL ) &&
+            ( pxContext->_internal.pxProcessFunction( pxContext,
+                                                      pxAzureIoTHubClient,
+                                                      ( void * ) pxPublishInfo ) == eAzureIoTHubClientSuccess ) )
+        {
+            break;
+        }
+    }
```
hihigupt, line 63: should we log if no function was able to handle the topic?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
...
+/*
+ * Topic subscribe state
+ */
+#define azureiothubTOPIC_SUBSCRIBE_STATE_NONE      ( 0x0 )
+#define azureiothubTOPIC_SUBSCRIBE_STATE_SUB       ( 0x1 )
+#define azureiothubTOPIC_SUBSCRIBE_STATE_SUBACK    ( 0x2 )
+
+/*-----------------------------------------------------------*/
+
+static void prvMQTTProcessIncomingPublish( AzureIoTHubClient_t * pxAzureIoTHubClient,
```
hihigupt, line 49: comment to function?

danewalton: yup just pushed a few updates

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
...
+                                                      pxAzureIoTHubClient,
+                                                      ( void * ) pxPublishInfo ) == eAzureIoTHubClientSuccess ) )
+        {
+            break;
+        }
+    }
+}
+/*-----------------------------------------------------------*/
+
+static void prvMQTTProcessResponse( AzureIoTHubClient_t * pxAzureIoTHubClient,
```
hihigupt, line 77: >prvMQTTProcessResponse [](start = 12, length = 22)

can we rename this processSubACK?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,544 @@
...
+                                    AzureIoTMQTTPacketInfo_t * pxIncomingPacket,
+                                    uint16_t usPacketId )
+{
+    uint32_t ulIndex;
+    AzureIoTHubClientReceiveContext_t * pxContext;
+
+    if( ( pxIncomingPacket->type & 0xF0U ) == AZURE_IOT_MQTT_PACKET_TYPE_SUBACK )
+    {
```
hihigupt, line 87: this is not required as line 104 does that check  .... you can add assert just to be sure.

source/azure_iot_hub_client.c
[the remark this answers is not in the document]
hihigupt: there should be no sub ack failure as iothub impl will break the tcp connection

---
In reply to: [607449368](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r607449368) [](ancestors = 607449368)

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+static void prvMQTTProcessSuback( AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                    AzureIoTMQTTPacketInfo_t * pxIncomingPacket,
+                                    uint16_t usPacketId )
+{
+    uint32_t ulIndex;
+    AzureIoTHubClientReceiveContext_t * pxContext;
+
+    configASSERT( pxIncomingPacket != NULL );
+
+    for( ulIndex = 0; ulIndex < azureiothubSUBSCRIBE_FEATURE_COUNT; ulIndex++ )
```
hihigupt, line 86: should we assert on packettype too?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    /* Failed means no topic match */
+    if( az_result_failed( az_iot_hub_client_c2d_parse_received_topic( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
+                                                                      xTopicSpan, &xOutEmbeddedRequest ) ) )
+    {
+        return eAzureIoTHubClientTopicNoMatch;
+    }
+
+    /* Process incoming Publish. */
+    AZLogInfo( ( "Cloud xCloudToDeviceMessage QoS : %d.", xMQTTPublishInfo->qos ) );
+
```
hihigupt, line 150: is this required?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    /* Verify the received publish is for the feature we have subscribed to. */
+    AZLogInfo( ( "Cloud xCloudToDeviceMessage topic: %.*s  with payload : %.*s.",
```
hihigupt, line 152: can this be moved to Debug level

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+
+    /* Failed means no topic match */
+    if( az_result_failed( az_iot_hub_client_methods_parse_received_topic( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
+                                                                          xTopicSpan, &xOutEmbeddedRequest ) ) )
+    {
+        return eAzureIoTHubClientTopicNoMatch;
+    }
+
+    /* Process incoming Publish. */
+    AZLogInfo( ( "Direct method QoS : %d.", xMQTTPublishInfo->qos ) );
```
hihigupt, line 193: >AZLogInfo [](start = 4, length = 9)

this can also be removed and the following log can be moved to debug level

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+/**
+ *
+ * Check/Process messages for incoming direct method messages.
+ *
+ * */
+static uint32_t prvAzureIoTHubClientDirectMethodProcess( AzureIoTHubClientReceiveContext_t * pxContext,
+                                                         AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                         void * pxPublishInfo )
+{
+    AzureIoTHubClientMethodRequest_t xMethodRequest;
```
hihigupt, line 180: >xMethodRequest [](start = 37, length = 14)

= { 0 }?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+/**
+ *
+ * Check/Process messages for incoming Cloud to Device messages.
+ *
+ * */
+static uint32_t prvAzureIoTHubClientC2DProcess( AzureIoTHubClientReceiveContext_t * pxContext,
+                                                AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                void * pxPublishInfo )
+{
+    AzureIoTHubClientCloudToDeviceMessageRequest_t xCloudToDeviceMessage;
```
hihigupt, line 136: = { 0 }

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+ * */
+static uint32_t prvAzureIoTHubClientDeviceTwinProcess( AzureIoTHubClientReceiveContext_t * pxContext,
+                                                       AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                       void * pxPublishInfo )
+{
+    AzureIoTHubClientTwinResponse_t xTwinResponse;
+    AzureIoTMQTTPublishInfo_t * xMQTTPublishInfo = ( AzureIoTMQTTPublishInfo_t * ) pxPublishInfo;
+    az_iot_hub_client_twin_response xOutRequest;
+    az_span xTopicSpan = az_span_create( ( uint8_t * ) xMQTTPublishInfo->pTopicName, xMQTTPublishInfo->topicNameLength );
+    uint32_t xRequestID = 0;
```
hihigupt, line 230: >xRequestID [](start = 13, length = 10)

ulRequestID

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+
+    /* Failed means no topic match */
+    if( az_result_failed( az_iot_hub_client_twin_parse_received_topic( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
+                                                                       xTopicSpan, &xOutRequest ) ) )
+    {
+        return eAzureIoTHubClientTopicNoMatch;
```
hihigupt, line 236: AZLogWarn( ( "Twin topic parsing failed: 0x%08x", xCoreResult ) );

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    /* Process incoming Publish. */
+    AZLogInfo( ( "Twin Incoming QoS : %d.", xMQTTPublishInfo->qos ) );
+
```
hihigupt, line 241: not required

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+        if( az_span_size( xOutRequest.xRequestID ) == 0 )
+        {
+            xTwinResponse.messageType = eAzureIoTHubTwinDesiredPropertyMessage;
+        }
+        else
+        {
+            if( az_result_failed( az_span_atou32( xOutRequest.xRequestID, &xRequestID ) ) )
+            {
+                return eAzureIoTHubClientFailed;
+            }
```
hihigupt, line 260: how about arranging it like:

static uint32_t prvAzureIoTHubClientDeviceTwinProcess( AzureIoTHubClientReceiveContext_t * pxContext,
                                                       AzureIoTHubClient_t * pxAzureIoTHubClient,
                                                       void * pxPublishInfo )
{
    AzureIoTHubClientTwinResponse_t xMessage;
    AzureIoTMQTTPublishInfo_t * xMqttPublishInfo = ( AzureIoTMQTTPublishInfo_t * ) pxPublishInfo;
    az_iot_hub_client_twin_response xOutRequest;
    az_span xTopic = az_span_create( ( uint8_t * ) xMqttPublishInfo->pcTopicName,
                                     xMqttPublishInfo->usTopicNameLength );
    az_result xCoreResult;
    AzureIoTHubClientResult_t xReturn;
    uint32_t ulRequestID = 0;

    xCoreResult = 
        az_iot_hub_client_twin_parse_received_topic( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
                                                     xTopic, &xOutRequest );
    if( az_result_failed( xCoreResult ) )
    {
        AZLogWarn( ( "Twin topic parsing failed: 0x%08x", xCoreResult ) );
        xReturn = eAzureIoTHubClientFailed;
    }
    else
    {
        /* Verify the received publish is for the we have subscribed to. */
        AZLogInfo( ( "Twin topic: %.*s. with payload : %.*s",
                     mqttPublishInfo->topicNameLength,
                     mqttPublishInfo->pTopicName,
                     mqttPublishInfo->payloadLength,
                     mqttPublishInfo->pPayload ) );

        xReturn = eAzureIoTHubClientSuccess;

        if( pxContext->_internal.callbacks.xTwinCallback )
        {
            if( az_span_size( xOutRequest.request_id ) == 0 )
            {
                xMessage.xMessageType = eAzureIoTHubTwinDesiredPropertyMessage;
            }
            else if( az_result_failed( xCoreResult = az_span_atou32( xOutRequest.request_id, &ulRequestID ) ) )
            {
                if( ulRequestID & 0x01 )
                {
                    xMessage.xMessageType = eAzureIoTHubTwinReportedResponseMessage;
                }
                else
                {
                    xMessage.xMessageType = eAzureIoTHubTwinGetMessage;
                }
            }
            else
            {
                /* Failed to parse the message */
                AZLogError( ( "Request ID parsing failed: 0x%08x", xCoreResult ) );
                xReturn = eAzureIoTHubClientFailed;
            }

            if( xReturn )
            {
                xMessage.pvMessagePayload = xMqttPublishInfo->pvPayload;
                xMessage.ulPayloadLength = xMqttPublishInfo->xPayloadLength;
                xMessage.xMessageStatus = xOutRequest.status;
                xMessage.pucVersion = az_span_ptr( xOutRequest.version );
                xMessage.usVersionLength = ( size_t ) az_span_size( xOutRequest.version );
                xMessage.ulRequestID = ulRequestID;
                pxContext->_internal.callbacks.xTwinCallback( &xMessage,
                                                              pxContext->_internal.pvCallbackContext );
            }
        }
    }

    return xReturn;
}

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    return eAzureIoTHubClientSuccess;
+}
+/*-----------------------------------------------------------*/
+
+/**
+ *
+ * Time callback for MQTT initialization.
+ *
+ * */
+static uint32_t prvGetTimeMs( void )
```
hihigupt, line 291: >void [](start = 30, length = 4)

do we do this? I just want to be consistent

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+{
+    TickType_t xTickCount = 0;
+    uint32_t ulTimeMs = 0UL;
```
hihigupt, line 294: >= 0UL [](start = 22, length = 5)

not required these initializations

hihigupt: how about just 

static uint32_t prvGetTimeMs( void )
{
return xTaskGetTickCount() * azureiotMILLISECONDS_PER_TICK;
}


---
In reply to: [609119426](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r609119426) [](ancestors = 609119426)

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    {
+        AZLogError( ( "Failed to initialize options: Invalid argument." ) );
+        xResult = eAzureIoTHubClientInvalidArgument;
+    }
+    else
+    {
+        pxHubClientOptions->pucModelID = NULL;
+        pxHubClientOptions->ulModelIDLength = 0;
+        pxHubClientOptions->pucModuleID = NULL;
+        pxHubClientOptions->ulModuleIDLength = 0;
```
hihigupt, line 320: memset?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+             ( ulBufferLength < ( azureiotUSERNAME_MAX + azureiotPASSWORD_MAX ) ) )
+    {
+        AZLogError( ( "Failed to initialize: Not enough memory passed." ) );
+        xResult = eAzureIoTHubClientOutOfMemory;
+    }
+    else
+    {
+        memset( ( void * ) pxAzureIoTHubClient, 0, sizeof( AzureIoTHubClient_t ) );
+
+        /* Setup scratch buffer to be used by middleware */
```
hihigupt, line 368: >scratch [](start = 17, length = 7)

Working buffer?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+            xHubOptions.module_id = az_span_create( ( uint8_t * ) pxHubClientOptions->pucModuleID, ( int32_t ) pxHubClientOptions->ulModuleIDLength );
+            xHubOptions.model_id = az_span_create( ( uint8_t * ) pxHubClientOptions->pucModelID, ( int32_t ) pxHubClientOptions->ulModelIDLength );
+            xHubOptions.user_agent = az_span_create( ( uint8_t * ) pxHubClientOptions->pucUserAgent, ( int32_t ) pxHubClientOptions->ulUserAgentLength );
+        }
+        else
+        {
+            xHubOptions.user_agent = az_span_create( ( uint8_t * ) azureiothubUSER_AGENT, sizeof( azureiothubUSER_AGENT ) - 1 );
+        }
+
+        if( az_result_failed( az_iot_hub_client_init( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
```
hihigupt, line 392: >az_result_failed [](start = 12, length = 16)

should we set and log embedded sdk result?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+                                                      xHostnameSpan, xDeviceIDSpan, &xHubOptions ) ) )
+        {
+            AZLogError( ( "Failed to initialize az_iot_hub_client_init." ) );
+            xResult = eAzureIoTHubClientInitFailed;
+        }
+        /* Initialize AzureIoTMQTT library. */
+        else if( ( xMQTTResult = AzureIoTMQTT_Init( &( pxAzureIoTHubClient->_internal.xMQTTContext ), pxTransportInterface,
+                                      prvGetTimeMs, prvEventCallback,
+                                      pucNetworkBuffer, ulNetworkBufferLength ) ) != AzureIoTMQTTSuccess )
```
hihigupt, line 401: lets align the code like 392

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+        {
+            AZLogError( ( "Failed to initialize AzureIoTMQTT_Init: %08x.", xMQTTResult ) );
```
hihigupt, line 403: >  [](start = 66, length = 1)

0x%08x

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+/*-----------------------------------------------------------*/
+
+void AzureIoTHubClient_Deinit( AzureIoTHubClient_t * pxAzureIoTHubClient )
+{
+    ( void ) pxAzureIoTHubClient;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTHubClientResult_t AzureIoTHubClient_Connect( AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                     bool cleanSession,
```
hihigupt, line 428: >cleanSession [](start = 58, length = 12)

xCleanSession

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+                                                     uint32_t ulTimeoutMilliseconds )
+{
+    AzureIoTMQTTConnectInfo_t xConnectInfo = { 0 };
+    AzureIoTHubClientResult_t xResult;
+    AzureIoTMQTTStatus_t xMQTTResult;
+    bool xSessionPresent;
+    uint32_t ulPasswordLength = 0;
+    size_t xMQTTUserNameLength;
+    az_result xEmbeddedResult;
```
hihigupt, line 437: >xEmbeddedResult [](start = 14, length = 15)

How about making it like xCoreResult,

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+                                                                                 ( char * ) xConnectInfo.pUserName, azureiotUSERNAME_MAX,
+                                                                                 &xMQTTUserNameLength ) ) )
+        {
+            AZLogError( ( "Failed to get username: %08x.", xEmbeddedResult ) );
+            xResult = eAzureIoTHubClientFailed;
+        }
+        /* Check if token refresh is set, then generate password */
+        else if( ( pxAzureIoTHubClient->_internal.pxAzureIoTHubClientTokenRefresh ) &&
+                 ( pxAzureIoTHubClient->_internal.pxAzureIoTHubClientTokenRefresh( pxAzureIoTHubClient,
+                                                                                   pxAzureIoTHubClient->_internal.xAzureIoTHubClientTimeFunction() + azureiothubDEFAULT_TOKEN_TIMEOUT_IN_SEC,
```
hihigupt, line 460: break this line with tab

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+                                                                                   pxAzureIoTHubClient->_internal.pucAzureIoTHubClientSymmetricKey,
+                                                                                   pxAzureIoTHubClient->_internal.ulAzureIoTHubClientSymmetricKeyLength,
+                                                                                   ( uint8_t * ) xConnectInfo.pPassword, azureiotPASSWORD_MAX,
+                                                                                   &ulPasswordLength ) ) )
+        {
+            AZLogError( ( "Failed to generate auth token." ) );
```
hihigupt, line 466: >auth [](start = 46, length = 4)

SAS token

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+            xResult = eAzureIoTHubClientFailed;
+        }
+        else
+        {
+            xConnectInfo.cleanSession = cleanSession;
```
hihigupt, line 471: >cleanSession [](start = 40, length = 12)

xConnection info might have changed its member name. Can you please check.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+
+            /* Send MQTT CONNECT packet to broker. Last Will and Testament is not used. */
+            if( ( xMQTTResult = AzureIoTMQTT_Connect( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
+                                                      &xConnectInfo,
+                                                      NULL,
+                                                      ulTimeoutMilliseconds,
+                                                      &xSessionPresent ) ) != AzureIoTMQTTSuccess )
+            {
+                AZLogError( ( "Failed to establish MQTT connection: Server=%.*s, MQTTStatus=%d.",
+                              pxAzureIoTHubClient->_internal.ulHostnameLength, pxAzureIoTHubClient->_internal.pucHostname,
```
hihigupt, line 486: cast like line 494

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+
+AzureIoTHubClientResult_t AzureIoTHubClient_Disconnect( AzureIoTHubClient_t * pxAzureIoTHubClient )
+{
+    AzureIoTMQTTStatus_t xMQTTResult;
+    AzureIoTHubClientResult_t xResult;
+
+    if( pxAzureIoTHubClient == NULL )
+    {
+        AZLogError( ( "Failed to disconnect: Invalid argument." ) );
+        return eAzureIoTHubClientInvalidArgument;
```
hihigupt, line 512: >return [](start = 8, length = 6)

how about like other API impl (line 442)  to assign xResult = eAzureIoTHubClientInvalidArgument; 
 and only return at end.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    if( ( xMQTTResult = AzureIoTMQTT_Disconnect( &( pxAzureIoTHubClient->_internal.xMQTTContext ) ) ) != AzureIoTMQTTSuccess )
+    {
+        AZLogError( ( "Failed to disconnect: %d.", xMQTTResult ) );
```
hihigupt, line 517: >%d [](start = 45, length = 2)

hex, like all other places?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+                                                                                                ( char * ) pxAzureIoTHubClient->_internal.pucAzureIoTHubClientWorkingBuffer,
+                                                                                                pxAzureIoTHubClient->_internal.ulAzureIoTHubClientWorkingBufferLength,
+                                                                                                &xTelemetryTopicLength ) ) )
+    {
+        AZLogError( ( "Failed to get telemetry topic : %08x.", xEmbeddedResult ) );
+        xResult = eAzureIoTHubClientFailed;
+    }
+    else
+    {
+        xMQTTPublishInfo.qos = AzureIoTMQTTQoS1;
```
hihigupt, line 559: >qos [](start = 25, length = 3)

these members has changed. Please update accordingly

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+        xMQTTPublishInfo.payloadLength = ulTelemetryDataLength;
+
+        /* Get a unique packet id. */
+        usPublishPacketIdentifier = AzureIoTMQTT_GetPacketId( &( pxAzureIoTHubClient->_internal.xMQTTContext ) );
+
+        /* Send PUBLISH packet. */
+        if( ( xMQTTResult = AzureIoTMQTT_Publish( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
+                                                  &xMQTTPublishInfo, usPublishPacketIdentifier ) ) != AzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to publish telementry : Error=%d.", xMQTTResult ) );
```
hihigupt, line 572: >d [](start = 65, length = 1)

hex

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+            xResult = eAzureIoTHubClientFailed;
```
hihigupt, line 573: >eAzureIoTHubClientFailed [](start = 22, length = 24)

I have added new error for publish.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    return xResult;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                           const uint8_t * pucTelemetryData,
+                                                           uint32_t ulTelemetryDataLength,
+                                                           AzureIoTMessageProperties_t * pxProperties )
```
hihigupt, line 534: @jspaith Should we add a timeout into this Api, and wait for ack till timeout ? This way we are consistent with out subscribe/unsubscribe api's

source/include/azure_iot_hub_client.h
```diff
@@ -21,7 +21,7 @@
 #include "azure_iot_mqtt_port.h"
 #include "azure_iot_transport_interface.h"
 
-#define hubclientSUBSCRIBE_FEATURE_COUNT    3
```
hihigupt, line 24: How about creating separate small PR for header. So, that we have less conflict

danewalton: yea I was going to but these changes were simple enough. All the other ones I've created issues for.

source/azure_iot_hub_client.c
danewalton: I think we should for functions with no args

source/azure_iot_hub_client.c
danewalton: should we do it like the embedded?

https://github.com/Azure/azure-sdk-for-c/blob/fed9dd0b2a000ab7e607bac76112a983f2bd8829/sdk/src/azure/iot/az_iot_hub_client.c#L27-L32

```c
AZ_NODISCARD az_iot_hub_client_options az_iot_hub_client_options_default()
{
  return (az_iot_hub_client_options){ .module_id = AZ_SPAN_EMPTY,
                                      .user_agent = client_sdk_version,
                                      .model_id = AZ_SPAN_EMPTY };
}
```

compiler then gives you warning about uninitialized members

source/azure_iot_hub_client.c
danewalton: uncrustified

source/azure_iot_hub_client.c
danewalton: yup will do

source/azure_iot_hub_client.c
danewalton: rebased and changed

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,538 @@
...
+                 mqttPublishInfo->pTopicName,
+                 mqttPublishInfo->payloadLength,
+                 mqttPublishInfo->pPayload ) );
+
+    if( context->_internal.callbacks.xCloudToDeviceMessageCallback )
+    {
+        message.pvMessagePayload = mqttPublishInfo->pPayload;
+        message.ulPayloadLength = (uint32_t)mqttPublishInfo->payloadLength;
+        message.xProperties._internal.properties = out_request.properties;
+        context->_internal.callbacks.xCloudToDeviceMessageCallback( &message,
```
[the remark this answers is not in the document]
danewalton: good idea

source/azure_iot_hub_client.c
danewalton: what do you mean here?

hihigupt: pxAzureIoTHubClient->_internal.xAzureIoTHubClientTimeFunction() +
<tab>azureiothubDEFAULT_TOKEN_TIMEOUT_IN_SEC,

source/azure_iot_hub_client.c
danewalton: pardon the extra uncrustify stuff here but here's the change:
https://github.com/Azure/azure-iot-middleware-freertos/pull/60/commits/3da1d16544476821cb4ac269743012f53a89335b

@hihigupt @jspaith

source/azure_iot_hub_client.c
danewalton: resolving https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r607389793 to have the discussion here

danewalton: seems the addition would be here and (in the future) all send API's. (Method response, twin get, etc).

hihigupt: other api uses QoS0, we do not need in that.

danewalton: got it so we would add probably a qos option and a timeout in case using qos1?

hihigupt: HOw about this.... We have macro in impl for QoS of Telemetry and then in send wait for timeout if QoS1 otherwise return immediately

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+            ( pxContext->_internal.pxProcessFunction( pxContext,
+                                                      pxAzureIoTHubClient,
+                                                      ( void * ) pxPublishInfo ) == eAzureIoTHubClientSuccess ) )
+        {
+            break;
+        }
+
+        /* If reached the end of the list and haven't found a context, log none found */
+        if( ulIndex == azureiothubSUBSCRIBE_FEATURE_COUNT - 1 )
+        {
```
hihigupt, line 71: how about checking index == azureiothubSUBSCRIBE_FEATURE_COUNT outside the for loop and then logging it?

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+        if( pxContext->_internal.usMqttSubPacketID == usPacketId )
+        {
+            /* TODO: inspect packet to see is ack was successful*/
+            pxContext->_internal.usState = azureiothubTOPIC_SUBSCRIBE_STATE_SUBACK;
+            break;
+        }
+
+        /* If reached the end of the list and haven't found a context, log none found */
+        if( ulIndex == azureiothubSUBSCRIBE_FEATURE_COUNT - 1 )
+        {
```
hihigupt, line 106: same

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+    configASSERT( ( pxIncomingPacket->ucType & 0xF0U ) == azureiotmqttPACKET_TYPE_SUBACK );
+
+    for( ulIndex = 0; ulIndex < azureiothubSUBSCRIBE_FEATURE_COUNT; ulIndex++ )
+    {
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ ulIndex ];
+
+        if( pxContext->_internal.usMqttSubPacketID == usPacketId )
+        {
+            /* TODO: inspect packet to see is ack was successful*/
+            pxContext->_internal.usState = azureiothubTOPIC_SUBSCRIBE_STATE_SUBACK;
```
hihigupt, line 100: log debug that we received ack for which receive queue

source/azure_iot_hub_client.c
hihigupt: this is more like warning 

---
In reply to: [610882301](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r610882301) [](ancestors = 610882301)

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+
+/**
+ *
+ * Check/Process messages for incoming Cloud to Device messages.
+ *
+ * */
+static uint32_t prvAzureIoTHubClientC2DProcess( AzureIoTHubClientReceiveContext_t * pxContext,
+                                                AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                void * pxPublishInfo )
+{
```
hihigupt, line 148: pvPublishInfo

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+/*-----------------------------------------------------------*/
+
+/**
+ *
+ * Check/Process messages for incoming direct method messages.
+ *
+ * */
+static uint32_t prvAzureIoTHubClientDirectMethodProcess( AzureIoTHubClientReceiveContext_t * pxContext,
+                                                         AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                         void * pxPublishInfo )
```
hihigupt, line 196: >pxPublishInfo [](start = 64, length = 13)

pv

hihigupt: all the places where we have void pointer

---
In reply to: [610886237](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r610886237) [](ancestors = 610886237)

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+    az_iot_hub_client_twin_response xOutRequest;
+    az_span xTopicSpan = az_span_create( ( uint8_t * ) xMQTTPublishInfo->pcTopicName, xMQTTPublishInfo->usTopicNameLength );
+    uint32_t ulRequestID = 0;
+
+    /* Failed means no topic match */
+    xCoreResult = az_iot_hub_client_twin_parse_received_topic( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
+                                                               xTopicSpan, &xOutRequest );
+
+    if( az_result_failed( xCoreResult ) )
+    {
```
hihigupt, line 264: looks much cleaner .... can we make all the process symmetric like this .

core = parse_api();

if (failed( core))
{
}
else
{
}

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,624 @@
...
+            xConnectInfo.usUserNameLength = ( uint16_t ) xMQTTUserNameLength;
+            xConnectInfo.usKeepAliveSeconds = azureiothubKEEP_ALIVE_TIMEOUT_SECONDS;
+            xConnectInfo.usPasswordLength = ( uint16_t ) ulPasswordLength;
+
+            /* Send MQTT CONNECT packet to broker. Last Will and Testament is not used. */
+            if( ( xMQTTResult = AzureIoTMQTT_Connect( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
+                                                      &xConnectInfo,
+                                                      NULL,
+                                                      ulTimeoutMilliseconds,
+                                                      &pxOutSessionPresent ) ) != eAzureIoTMQTTSuccess )
```
hihigupt, line 526: >pxOutSessionPresent [](start = 55, length = 19)

not sure why you need double pointer

hihigupt: and type is uint8_t

---
In reply to: [610890192](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r610890192) [](ancestors = 610890192)

source/azure_iot_hub_client.c
hihigupt: I prefer not to pass structs by return values.

source/azure_iot_hub_client.c
danewalton: oops yea find all and replace error here

source/azure_iot_hub_client.c
danewalton: Tracking
https://github.com/Azure/azure-iot-middleware-freertos/issues/80

source/azure_iot_hub_client.c
hihigupt: this seems to be like Debug level logging not info.

---
In reply to: [610883286](https://github.com/Azure/azure-iot-middleware-freertos/pull/60#discussion_r610883286) [](ancestors = 610883286)

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+            AZLogError( ( "Failed to initialize AzureIoTMQTT_Init: 0x%08x.", xMQTTResult ) );
+            xResult = eAzureIoTHubClientInitFailed;
+        }
+        else
+        {
+            pxAzureIoTHubClient->_internal.pucDeviceID = pucDeviceId;
+            pxAzureIoTHubClient->_internal.ulDeviceIDLength = ulDeviceIdLength;
+            pxAzureIoTHubClient->_internal.pucHostname = pucHostname;
+            pxAzureIoTHubClient->_internal.ulHostnameLength = ulHostnameLength;
+            pxAzureIoTHubClient->_internal.xAzureIoTHubClientTimeFunction = xGetTimeFunction;
```
hihigupt, line 460: >xAzureIoTHubClientTimeFunction [](start = 43, length = 30)

nit: azureIoTHubclient is redundent in the name, possibly we can make it smaller like xGetTimeFunction

hihigupt (approved):
:shipit:

source/azure_iot_hub_client.c
danewalton: we would probably have to change all the others like HMAC, token refresh, etc for parity sake

hihigupt: Sure, but something like this:
https://github.com/Azure/azure-iot-middleware-freertos/blob/4ad026331a8b5c9642fb9f8bc91cf25b1af50a9c/source/include/azure_iot_provisioning_client.h#L75

danewalton: gotcha gotcha. done

jspaith (approved):

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+                                                      ( void * ) pxPublishInfo ) == eAzureIoTHubClientSuccess ) )
+        {
+            break;
+        }
+    }
+
+    /* If reached the end of the list and haven't found a context, log none found */
+    if( ulIndex == azureiothubSUBSCRIBE_FEATURE_COUNT )
+    {
+        AZLogInfo( ( "No receive context found for incoming publish." ) );
```
jspaith, line 73: Consider logging topic this came on?  If I start seeing bunches of these in a long my first question would be "what is triggering this in 1st place."

Similar feedback for SUBACK processing.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+    configASSERT( pxIncomingPacket != NULL );
+    configASSERT( ( pxIncomingPacket->ucType & 0xF0U ) == azureiotmqttPACKET_TYPE_SUBACK );
+
+    for( ulIndex = 0; ulIndex < azureiothubSUBSCRIBE_FEATURE_COUNT; ulIndex++ )
+    {
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ ulIndex ];
+
+        if( pxContext->_internal.usMqttSubPacketID == usPacketId )
+        {
+            /* TODO: inspect packet to see is ack was successful*/
```
jspaith, line 99: Is there a GitHub tracking this?  If no, please create per my fear of TODOs.  (Back in the day we called this BUGBUG and we ended up with tons of them which is why I keep raising this).  Also if/when there's a GitHub, we should link to the issue from the code itself to help with background.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+    az_iot_hub_client_c2d_request xOutEmbeddedRequest;
+    az_span xTopicSpan = az_span_create( ( uint8_t * ) xMQTTPublishInfo->pcTopicName, xMQTTPublishInfo->usTopicNameLength );
+
+    /* Failed means no topic match */
+    xCoreResult = az_iot_hub_client_c2d_parse_received_topic( &pxAzureIoTHubClient->_internal.xAzureIoTHubClientCore,
+                                                              xTopicSpan, &xOutEmbeddedRequest );
+
+    if( az_result_failed( xCoreResult ) )
+    {
+        AZLogWarn( ( "Cloud to device topic parsing failed: 0x%08x", xCoreResult ) );
```
jspaith, line 163: I don't think this should be a warning - i don't think it should be logged at all, in fact.  This is going to be part of normal operation for say a deviceMethod invocation, and these warnings would cause a thoughtful developer to spend time trying to understand what was going wrong.

jspaith: I'd put in this block (and others) a brief note about how embedded API works because the topicNotMatch failure may not be obvious to someone debugging something that's not familiar with embedded layer.

Maybe something like in the comment right above embedded to expand a bit:
/* Failed means no topic match.  This is not a fatal error, but only means this message wasn't intended for C2D. */

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+        if( pxContext->_internal.callbacks.xMethodCallback )
+        {
+            xMethodRequest.pvMessagePayload = xMQTTPublishInfo->pvPayload;
+            xMethodRequest.ulPayloadLength = ( uint32_t ) xMQTTPublishInfo->xPayloadLength;
+            xMethodRequest.pucMethodName = az_span_ptr( xOutEmbeddedRequest.name );
+            xMethodRequest.usMethodNameLength = ( uint16_t ) az_span_size( xOutEmbeddedRequest.name );
+            xMethodRequest.pucRequestID = az_span_ptr( xOutEmbeddedRequest.request_id );
+            xMethodRequest.usRequestIDLength = ( uint16_t ) az_span_size( xOutEmbeddedRequest.request_id );
+
+            AZLogDebug( ( "Invoking Methods Callback." ) );
```
jspaith, line 235: Nit: I think this should be "Method Callback" and not "Methods Callback."

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+                                                               xTopicSpan, &xOutRequest );
+
+    if( az_result_failed( xCoreResult ) )
+    {
+        AZLogWarn( ( "Twin topic parsing failed: 0x%08x", xCoreResult ) );
+        xResult = eAzureIoTHubClientTopicNoMatch;
+    }
+    else
+    {
+        /* Verify the received publish is for the feature we have subscribed to. */
```
jspaith, line 275: I think this comment is either unneeded and/or in wrong place?  We've already verified (past tense) it's right publish for feature by the time we've got into this block.

If we extend the comments about about what failing the az_embedded_parse_whatever_topic() failures signifying not topic match, I think we don't need this comment (and analogous ones other places).

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+                    else
+                    {
+                        xTwinResponse.xMessageType = eAzureIoTHubTwinGetMessage;
+                    }
+                }
+                else
+                {
+                    /* Failed to parse the message */
+                    AZLogError( ( "Request ID parsing failed: 0x%08x", xCoreResult ) );
+                    xResult = eAzureIoTHubClientFailed;
```
jspaith, line 307: I understand that we need to set some signal if we get here so as not to invoke the callback.  However we don't want to return this to the caller in the PUBLISH processing logic, because it will interpret that as a TopicNotMatch and continue through other topics where really caller should fail.

source/azure_iot_hub_client.c
```diff
@@ -0,0 +1,630 @@
...
+        /* Send PUBLISH packet. */
+        if( ( xMQTTResult = AzureIoTMQTT_Publish( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
+                                                  &xMQTTPublishInfo, usPublishPacketIdentifier ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to publish telementry: 0x%08x.", xMQTTResult ) );
+            xResult = eAzureIoTHubClientPublishFailed;
+        }
+        else
+        {
+            xResult = eAzureIoTHubClientSuccess;
```
jspaith, line 625: Nit: Other places after going to AzureIoTMQTT_* layer on success, there's a LogInfo().  I think one should be here too.

jspaith (approved):

source/azure_iot_hub_client.c
danewalton: good call. done

danewalton: note the suback doesn't have that kind of information. only a packet id.

source/azure_iot_hub_client.c
danewalton: yup here:
https://github.com/Azure/azure-iot-middleware-freertos/issues/65
might not be needed though will revisit when attending to the issues

source/azure_iot_hub_client.c
danewalton: Yes in the full version there is a "process" function devoted to a specific subscription. So this embedded "parse" is supposed to only get the information out not necessarily check that it is for this feature. If this fails, it means somehow a topic got through which isn't for this subscibe.
https://github.com/Azure/azure-iot-middleware-freertos/blob/98b041e3cbf874a47e684451397f8603c8503a39/source/azure_iot_hub_client.c#L829

Will add the comment clarification

source/azure_iot_hub_client.c
danewalton: yea this is interesting. I worry that some error in a process function for a feature it isn't for could prevent it from getting processed by the correct function if we bail out on error. Not sure how recoverable we want this to be.

jspaith: @danewalton - tagging just since this got merged (congrats!).

I actually don't think this function should be returning an ERROR.  The caller doesn't care whether there was an error here or not.

This function (and friends) should be returning like a `bool` for "I claim ownership of this topic" or "I matched this topic."  And from caller perspective it's basically saying "OK, the topic matched, stop going through loop."  No matter what happens after the topic match, since we can only match one topic the caller should stop processing.

PR #62, opened 2021-04-05
Add provisioning unit tests

hihigupt:


danewalton (approved):

tests/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
```
jspaith, line 4: The last I checked, Ubuntu 18.04 only supports cmake version 3.10.  Since that's probably decent % of Linux dev out there, what do we complicate if we limit ourselves to 3.10 features?

tests/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+add_link_options(-fprofile-arcs -ftest-coverage)
+
+# FreeRTOS Kernel include
+include_directories(${AZ_IOT_REPO_ROOT}/libraries/FreeRTOS/FreeRTOS/Source/include)
+
+# port for FreeRTOS
+if(UNIX)
+    include_directories(${AZ_IOT_REPO_ROOT}/libraries/FreeRTOS/FreeRTOS/Source/portable/ThirdParty/GCC/Posix)
+else()
+    message( FATAL_ERROR "Tests only supported in linux." )
```
jspaith, line 23: Is the requirement making this Linux only the code coverage or are there other problems?  

UT should in theory be able to be run anywhere.  And I'm very attached to windbg :).

tests/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+endif()
+
+# Middleware source
+add_library(az_iot_middleware_freertos
+  ${AZ_IOT_REPO_ROOT}/source/azure_iot_provisioning_client.c
+)
+
+target_compile_options(az_iot_middleware_freertos
+    PRIVATE
+    -Werror
```
jspaith, line 33: Nit: These feel like valuable flags for building anything, including samples.  (Our old practice of letting samples compile without as many warnings we should get away from.)

Could we make these global somehow?  That'd also help for like letting this build on other platforms, so in some global file
if (GCC) {
  set(COMPILER_WARNINGS -Werror -Wall etc.)
)
else if (MSVC) {
  set(COMPILER_WARNINGS -whateverMSBuildUses
)

tests/config_files/FreeRTOSConfig.h
```diff
@@ -0,0 +1,180 @@
...
+
+/* Ensure stdint is only used by the compiler, and not the assembler. */
+#if defined( __ICCARM__ ) || defined( __CC_ARM ) || defined( __GNUC__ )
+    #include <stdint.h>
+    extern uint32_t SystemCoreClock;
+#endif
+
+#define configSUPPORT_STATIC_ALLOCATION              1
+
+#define configUSE_PREEMPTION                         1
```
jspaith, line 50: Do we need all these FreeRTOS specific config settings, for UT at least?  I guess it's OK if we do (I'm assuming one of our proudct files sucks in FreeRTOS.h which wants all this?) but in perfect world there's a bit more abstraction from OS.

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+}
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTProvisioningClient_SymmetricKeySet_Failure( void ** ppvState )
+{
+    AzureIoTProvisioningClient_t xTestProvisioningClient;
+
+    ( void ) ppvState;
+
+    /* Fail AzureIoTProvisioningClient_SetSymmetricKey when null symmetric key is passed */
```
jspaith, line 347: Nit: Shouldn't there also be a test when `xTestProvisioningClient==NULL` and also a non-NULL `ucSymmetricKey` but `0` for its length?

jspaith: Similar feedback other places like `AzureIoTProvisioningClient_Register` fail tests where param validation isn't done for handle.

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+{
+    AzureIoTProvisioningClient_t xTestProvisioningClient;
+    AzureIoTProvisioningClientResult_t xResult;
+
+    ( void ) ppvState;
+
+    prvSetupTestProvisioningClient( &xTestProvisioningClient );
+
+    /* Connect */
+    will_return( prvHmacFunction, 0 );
```
jspaith, line 405: Nit: Consider making these 4 lines of doing the will_return to Register call into a helper function, since it's called a bunch and if anything changes in the implementation it'll be easier to change 1 helper that bunches of copy/paste.

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+        cmocka_unit_test( testAzureIoTProvisioningClient_Deinit_Success ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_SymmetricKeySet_Failure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_SymmetricKeySet_Success ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_ConnectFailure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_SubscribeFailure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_SubscribeAckFailure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_PublishFailure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_RegistrationFailure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_QueryFailure ),
+        cmocka_unit_test( testAzureIoTProvisioningClient_Register_Success )
```
jspaith, line 700: I'm assuming stuff like `AzureIoTProvisioningClient_GetDeviceAndHub` is being tested in a later PR?  If so that's actually great since it keeps this PR smaller but want to make sure.?

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+    AzureIoTProvisioningClientResult_t xResult;
+
+    ( void ) ppvState;
+
+    prvSetupTestProvisioningClient( &xTestProvisioningClient );
+
+    /* Connect */
+    will_return( prvHmacFunction, 0 );
+    will_return( AzureIoTMQTT_Connect, eAzureIoTMQTTServerRefused );
+    will_return( AzureIoTMQTT_ProcessLoop, eAzureIoTMQTTSuccess );
```
jspaith, line 386: I didn't pick up on this in product code review, but if the _Connect call fails should we even try to do a ProcessLoop?  It seems we should bail out since I can't see how Process would succeed. ?

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+    xResult = AzureIoTProvisioningClient_Register( &xTestProvisioningClient, 0 );
+    assert_int_not_equal( xResult, eAzureIoTProvisioningPending );
```
jspaith, line 388: Could we just check against the expected error code directly?  I get there's some flexibility here, but it's a bit cleaner since it's one line and not 2 for check, and we tend to pin the behavior a bit better (so if we start returning eOutOfMemory where we'd always been eFailure before from product, that'd be good to know).

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+    xPacketInfo.ucType = 0;
+    xResult = AzureIoTProvisioningClient_Register( &xTestProvisioningClient, 0 );
+    assert_int_not_equal( xResult, eAzureIoTProvisioningPending );
+    assert_int_not_equal( xResult, eAzureIoTProvisioningSuccess );
+
+    AzureIoTProvisioningClient_Deinit( &xTestProvisioningClient );
+}
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTProvisioningClient_Register_Success( void ** ppvState )
```
jspaith, line 612: OK if on later PR, but could we add a test here where we the Register passed in a long timeout.  I'm interested in getting some test over `prvProvClientRunWorkflow` path where it kicks into its do{}while loop >1 time and with permutations like the `(ulTimeoutMilliseconds% azureiotprovisioningPROCESS_LOOP_TIMEOUT_MS) = 0` and where say `(ulTimeoutMilliseconds% azureiotprovisioningPROCESS_LOOP_TIMEOUT_MS) != 0`. 

This would lend itself to helper functions for these various success cases.

tests/azure_iot_cmocka_mqtt.c
```diff
@@ -0,0 +1,142 @@
...
+    AzureIoTMQTTResult_t xReturn = ( AzureIoTMQTTResult_t ) mock();
+
+    if( xReturn )
+    {
+        return xReturn;
+    }
+
+    if( pucPublishPayload )
+    {
+        assert_memory_equal( pxPublishInfo->pvPayload, pucPublishPayload, pxPublishInfo->xPayloadLength );
```
jspaith, line 89: Love being able to check the outgoing payload.  Do we use this anywhere in the actual test - I'm not seeing `pucPublishPayload` referenced in the UT except in the `external`.

tests/azure_iot_provisioning_client_ut.c
```diff
@@ -0,0 +1,704 @@
...
+    \"operationId\":\"4.002305f54fc89692.b1f11200-8776-4b5d-867b-dc21c4b59c12\",\"status\":\"assigned\",\"registrationState\": \
+         {\"registrationId\":\"reg_id\",\"createdDateTimeUtc\":\"2019-12-27T19:51:41.6630592Z\",\"assignedHub\":\"test.azure-iothub.com\", \
+          \"deviceId\":\"testId\",\"status\":\"assigned\",\"substatus\":\"initialAssignment\",\"lastUpdatedDateTimeUtc\":\"2019-12-27T19:51:41.8579703Z\", \
+          \"etag\":\"XXXXXXXXXXX=\"\
+         }\
+}";
+static const uint8_t ucAssigningHubResponse[] = "{ \
+    \"operationId\":\"4.002305f54fc89692.b1f11200-8776-4b5d-867b-dc21c4b59c12\",\"status\":\"assigning\" \
+}";
+static const uint8_t ucFailureHubResponse[] = "{ \
```
jspaith, line 46: Could we add additional bad responses in JSON blobs - from horrible stuff like not even JSON in 1st place to something that's close to right JSON but that maybe missing field or two that either embedded SDK or potentially middleware layer would be unhappy on missing?

jspaith (approved):

tests/azure_iot_provisioning_client_ut.c
hihigupt: sure.

---
In reply to: [607957371](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607957371) [](ancestors = 607957371)

tests/azure_iot_provisioning_client_ut.c
hihigupt: sure

---
In reply to: [607966442](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607966442) [](ancestors = 607966442)

.github/workflows/ci_tests_linux.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Submodules Init
+      uses: snickerbockers/submodules-init@v4
+    - name: Configure
+      run: .github/scripts/install_softwares.sh
+    - name: Unit Tests
+      run: cd tests ; cmake -Bbuild .; cmake --build build; cd build; ctest -T test
+    - name: Code Coverage
+      run: cd tests/build; gcovr -r $(pwd) -f ../../source/*.c
```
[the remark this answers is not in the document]
hihigupt: Yes, but lets get this feature in different PR. I will create issue about this.

---
In reply to: [607928533](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607928533) [](ancestors = 607928533)

tests/CMakeLists.txt
hihigupt: Its more like cmake support that is required. More robust solution is what @danewalton is working on. For now to test PR and unit test test running, I am using this cmake.

---
In reply to: [607944775](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607944775) [](ancestors = 607944775)

tests/CMakeLists.txt
hihigupt: I will defer it for now as @danewalton is working on cmake

---
In reply to: [607946369](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607946369) [](ancestors = 607946369)

tests/azure_iot_provisioning_client_ut.c
hihigupt: Yes, right now current impl trigger Action and then we always  process loop irrespective of action failed or not. 

I can update the impl.

---
In reply to: [607965358](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607965358) [](ancestors = 607965358)

tests/azure_iot_provisioning_client_ut.c
hihigupt: Yup

---
In reply to: [607973225](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607973225) [](ancestors = 607973225)

tests/azure_iot_provisioning_client_ut.c
hihigupt: Yes I will create new PR. Right now code coverage is 76, which is something I am not happy with. I will add additional tests to raise this code coverage.

---
In reply to: [607961133](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r607961133) [](ancestors = 607961133)

tests/azure_iot_cmocka_mqtt.c
hihigupt: I was using these in iothub tests.

tests/azure_iot_provisioning_client_ut.c
hihigupt: parsing is done by embedded sdk, and they already have these tests.. Just wondering if this will add value. But in the next PR to increase the line coverage I will add additional tests.

---
In reply to: [608075635](https://github.com/Azure/azure-iot-middleware-freertos/pull/62#discussion_r608075635) [](ancestors = 608075635)

jspaith: I think we should add at least one test to make the embedded SDK fail.  This will test that if we get bad JSON & embedded fails, then our layer also properly handles the scenario.

hihigupt: Sure, I will add in next PR.

tests/CMakeLists.txt
danewalton: freertos and it's libs (coremqtt etc) require 3.13. So we are essentially staying up to date with those

tests/CMakeLists.txt
danewalton: yea these will be update later in another PR. I didn't comment on these since these are being tackled elsewhere. This is ad hoc for now

PR #75, opened 2021-04-07
Add implmentation of IoTHub C2D/Methods/Twin

hihigupt:


hihigupt:
https://github.com/Azure/azure-iot-middleware-freertos/issues/82

source/azure_iot_hub_client.c
```diff
@@ -626,3 +724,503 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    else
+    {
+        xReturn = eAzureIoTHubClientSuccess;
+    }
+
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTHubClientResult_t AzureIoTHubClient_SubscribeCloudToDeviceMessage( AzureIoTHubClient_t * pxAzureIoTHubClient,
```
jspaith, line 756: I want to talk through the disconnect scenarios esp. with SUBSCRIBE.  I think code we have is fine but let's make sure my mental model is correct.

So we do Connect/SubscribeC2D (e.g.).  And then we disconnect.  So nothing's happening.

Then assume app does a AzIoT_Connect again and it succeeds.

If App is using session present and hub accepts, then nothing has NULL'd out `pxContext->_internal.pxProcessFunction` so we're still in business to deal with callbacks.

If app isn't using session present OR hub accepts connect but not new session, then app would be on its own to re-subscribe which would reset all these guys.

I know this is a long comment indicating I'm happy, but if I'm misreading this I need to be educated.  These retry issues can be bug farms.

source/azure_iot_hub_client.c
```diff
@@ -626,3 +724,503 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    if( ( pxAzureIoTHubClient == NULL ) ||
+        ( xCallback == NULL ) )
+    {
+        AZLogError( ( "Failed to subscribe cloud message : Invalid argument" ) );
+        xReturn = eAzureIoTHubClientInvalidArgument;
+    }
+    else
+    {
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ azureiothubRECEIVE_CONTEXT_INDEX_C2D ];
+        xMqttSubscription.xQoS = eAzureIoTMQTTQoS1;
```
jspaith, line 776: Dumb question and maybe answered in specs but why do we have Qos1 here but Qos0 on other subscriptions?

source/azure_iot_hub_client.c
```diff
@@ -626,3 +724,503 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+        xMqttSubscription.pcTopicFilter = ( const uint8_t * ) AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC;
+        xMqttSubscription.usTopicFilterLength = ( uint16_t ) sizeof( AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) - 1;
+        usSubscribePacketIdentifier = AzureIoTMQTT_GetPacketId( &( pxAzureIoTHubClient->_internal.xMQTTContext ) );
+
+        AZLogDebug( ( "Attemptting to unsubscribe from the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
+
+        if( ( xMQTTResult = AzureIoTMQTT_Unsubscribe( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
```
jspaith, line 827: It appears this is a fire and forget, and we mark ourselves as unsubscribed before an UNSUBACK returns correct?

I like this simplicity, but are there scenarios that this will cause problems?  I'm not trying to make us add more code here assuming we don't need...

Suppose for instance app does SubscribeC2D / UnsubscribeC2D.  And suppose we send the bytes so this function succeeds, but server isn't happy for whatever reason.  And then app comes in to SubscribeC2D again.  Since they're already subscribed as far as server is concerned, would MQTT let the 2nd subscription to same topic just be "cool" or would there be an error?

source/azure_iot_hub_client.c
```diff
@@ -346,6 +355,95 @@ static uint32_t prvGetTimeMs( void )
...
+        }
+
+        xReturn = eAzureIoTHubClientSuccess;
+    }
+
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+static AzureIoTHubClientResult_t prvWaitForSubAck( AzureIoTHubClient_t * pxAzureIoTHubClient,
```
jspaith, line 405: Nit - consider a LogDebug (I think that's right level) for when we enter and if we leave this function successfully.  Reason I suggest is that since we could conceivable be blocking for a bit during a SUBACK (as these are rare and we force caller to wait here, which is fine) there could be questions about a thread's lifeliness if we're gone "too long".  

I don't think we need/want per-ProcessLoop() as that might get to chatty (or that's some crazy packet level thing whic is probably diferent discussion).  Just the entry/exit.

source/azure_iot_hub_client.c
```diff
@@ -626,3 +724,503 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+        }
+        else
+        {
+            xMQTTPublishInfo.xQOS = eAzureIoTMQTTQoS0;
+            xMQTTPublishInfo.pcTopicName = pxAzureIoTHubClient->_internal.pucWorkingBuffer;
+            xMQTTPublishInfo.usTopicNameLength = ( uint16_t ) xTopicLength;
+            xMQTTPublishInfo.pvPayload = ( const void * ) pucReportedPayload;
+            xMQTTPublishInfo.xPayloadLength = ulReportedPayloadLength;
+
+            if( ( xMQTTResult = AzureIoTMQTT_Publish( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
```
jspaith, line 1150: Why are we using 0 for `usPacketId` here but in telemetry send we explicitly get a packetID?  Is it because of the Qos1 needing but not Qos0 (which makes sense as there's nothing to ACK, but I still want to at least double check my knowledge here).

source/azure_iot_hub_client.c
```diff
@@ -626,3 +724,503 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+            xReturn = eAzureIoTHubClientFailed;
+        }
+        else
+        {
+            pxContext->_internal.usState = azureiothubTOPIC_SUBSCRIBE_STATE_SUB;
+            pxContext->_internal.usMqttSubPacketID = usSubscribePacketIdentifier;
+            pxContext->_internal.pxProcessFunction = prvAzureIoTHubClientC2DProcess;
+            pxContext->_internal.callbacks.xCloudToDeviceMessageCallback = xCallback;
+            pxContext->_internal.pvCallbackContext = prvCallbackContext;
+            xReturn = prvWaitForSubAck( pxAzureIoTHubClient, pxContext, ulTimeoutMilliseconds );
```
jspaith, line 796: Should we be filling out the fields above before `prvWaitForSuback` returns successfully.  I worry there's a race where if doing ProcessLoop() in the wait, we get the SUBSCRIBE and also a PUBLISH to something app has SUB'd for in same packet then we could potentially end up invoking the application's callback *before* we returned from `AzureIoTHubClient_SubscribeCloudToDeviceMessage`.

I also would worry that if `xReturn` is a failure, we should leave the fields and callback null'd out.

source/azure_iot_hub_client.c
```diff
@@ -626,3 +724,503 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    return xReturn;
+}
+/*-----------------------------------------------------------*/
+
+AzureIoTHubClientResult_t AzureIoTHubClient_GetDeviceTwin( AzureIoTHubClient_t * pxAzureIoTHubClient )
+{
+    AzureIoTMQTTResult_t xMQTTResult;
+    AzureIoTHubClientResult_t xReturn;
+    AzureIoTMQTTPublishInfo_t xMQTTPublishInfo = { 0 };
+    uint8_t ucRequestID[ 10 ];
```
jspaith, line 1172: nit - magic number in code.

source/azure_iot_hub_client.c
```diff
@@ -346,6 +355,95 @@ static uint32_t prvGetTimeMs( void )
...
+    xCoreResult = az_span_u32toa( xSpan, pxAzureIoTHubClient->_internal.ulCurrentRequestID, &xRemainder );
+
+    if( az_result_failed( xCoreResult ) )
+    {
+        AZLogError( ( "Couldn't convert request id to span" ) );
+        xReturn = eAzureIoTHubClientFailed;
+    }
+    else
+    {
+        *pxSpan = az_span_slice( *pxSpan, 0, az_span_size( *pxSpan ) - az_span_size( xRemainder ) );
```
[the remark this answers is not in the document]
hihigupt: Any [in] paramerter of span generally is passed as copy and out is pointer, which is waste of stack space in [inout] parameter. Few api's of span and their usage:

API: [https://github.com/Azure/azure-sdk-for-c/blob/5d8baab842e36e277aa58fdb7ad1f287f1375e4d/sdk/inc/azure/core/az_span.h#L412](https://github.com/Azure/azure-sdk-for-c/blob/5d8baab842e36e277aa58fdb7ad1f287f1375e4d/sdk/inc/azure/core/az_span.h#L412)
usage:[https://github.com/Azure/azure-sdk-for-c/blob/6a61fdc180cb43d9bb6e22d06166724bd4156e73/sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino#L283](https://github.com/Azure/azure-sdk-for-c/blob/6a61fdc180cb43d9bb6e22d06166724bd4156e73/sdk/samples/iot/aziot_esp8266/aziot_esp8266.ino#L283)


---
In reply to: [612734661](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612734661) [](ancestors = 612734661)

source/azure_iot_hub_client.c
hihigupt: Yup, method response, reported property and get twin are QoS0, so no need of packet identifier.

---
In reply to: [612750989](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612750989) [](ancestors = 612750989)

source/azure_iot_hub_client.c
hihigupt: Ack

---
In reply to: [612753409](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612753409) [](ancestors = 612753409)

source/azure_iot_hub_client.c
hihigupt: Yes, to keep the logic simple I am sending the unsubscribe and not waiting for it... Reason I am waiting for suback in subscribe path is: User may send QoS0 twin message and my never receive the response as topic was not sub ack before sending the GetTwin request.

Yup subscribing an existing subscribed topic is ok... I verfied it by sending multiple requests.

---
In reply to: [612745816](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612745816) [](ancestors = 612745816)

source/azure_iot_hub_client.c
hihigupt: Because C2D messages are subscribed with QoS1, where as all other are fire and forget type (Request/response)

[https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#receiving-cloud-to-device-messages](https://docs.microsoft.com/en-us/azure/iot-hub/iot-hub-mqtt-support#receiving-cloud-to-device-messages)

When a device app subscribes to a topic with QoS 2, IoT Hub grants maximum QoS level 1 in the SUBACK packet. After that, IoT Hub delivers messages to the device using QoS 1.

---
In reply to: [612740875](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612740875) [](ancestors = 612740875)

source/azure_iot_hub_client.c
danewalton: oh interesting. So essentially what is the behavior of cloud to device if we have a disconnect then connect again with session present?

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    AzureIoTMQTTSubscribeInfo_t xMqttSubscription = { 0 };
+    AzureIoTMQTTResult_t xMQTTResult;
+    AzureIoTHubClientResult_t xReturn;
+    uint16_t usSubscribePacketIdentifier;
+    AzureIoTHubClientReceiveContext_t * pxContext;
+
+    if( ( pxAzureIoTHubClient == NULL ) ||
+        ( xCallback == NULL ) )
+    {
+        AZLogError( ( "Failed to subscribe cloud message : Invalid argument" ) );
```
danewalton, line 786:
```suggestion
        AZLogError( ( "Failed to subscribe cloud to device message : Invalid argument" ) );
```

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    }
+    else
+    {
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ azureiothubRECEIVE_CONTEXT_INDEX_C2D ];
+        xMqttSubscription.xQoS = eAzureIoTMQTTQoS1;
+        xMqttSubscription.pcTopicFilter = ( const uint8_t * ) AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC;
+        xMqttSubscription.usTopicFilterLength = ( uint16_t ) sizeof( AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) - 1;
+        usSubscribePacketIdentifier = AzureIoTMQTT_GetPacketId( &( pxAzureIoTHubClient->_internal.xMQTTContext ) );
+
+        AZLogDebug( ( "Attemptting to subscribe to the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
```
danewalton, line 797:
```suggestion
        AZLogDebug( ( "Attempting to subscribe to the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
```

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+
+        if( ( xMQTTResult = AzureIoTMQTT_Subscribe( &( pxAzureIoTHubClient->_internal.xMQTTContext ),
+                                                    &xMqttSubscription, 1, usSubscribePacketIdentifier ) ) != eAzureIoTMQTTSuccess )
+        {
+            AZLogError( ( "Failed to subscribe, error=0x%08x", xMQTTResult ) );
```
danewalton, line 802: just to unify sort of
```suggestion
            AZLogError( ( "Cloud to device subscribe failed: error=0x%08x", xMQTTResult ) );
```

source/azure_iot_hub_client.c
hihigupt: Yes, correct. Client when connect need to inspect session present or not (assuming clean session is pass false). If session is present on server, client need to set its state. Client can set its state if not set previously (e.g If done init on iothub client, client need to subscribe topic to set its state, otherwise no subscribe is required )

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+            xReturn = eAzureIoTHubClientFailed;
+        }
+        else
+        {
+            pxContext->_internal.usState = azureiothubTOPIC_SUBSCRIBE_STATE_SUB;
+            pxContext->_internal.usMqttSubPacketID = usSubscribePacketIdentifier;
+
+            if( ( xReturn = prvWaitForSubAck( pxAzureIoTHubClient, pxContext, ulTimeoutMilliseconds ) ) )
```
danewalton, line 810: we should probably have explicit equality/non equality checks and not assume what number success/failure is
```suggestion
            if( ( xReturn = prvWaitForSubAck( pxAzureIoTHubClient, pxContext, ulTimeoutMilliseconds ) )  != eAzureIoTMQTTSuccess )
```

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    }
+    else
+    {
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ azureiothubRECEIVE_CONTEXT_INDEX_C2D ];
+        xMqttSubscription.xQoS = eAzureIoTMQTTQoS1;
+        xMqttSubscription.pcTopicFilter = ( const uint8_t * ) AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC;
+        xMqttSubscription.usTopicFilterLength = ( uint16_t ) sizeof( AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) - 1;
+        usSubscribePacketIdentifier = AzureIoTMQTT_GetPacketId( &( pxAzureIoTHubClient->_internal.xMQTTContext ) );
+
+        AZLogDebug( ( "Attemptting to unsubscribe from the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
```
danewalton, line 849:
```suggestion
        AZLogDebug( ( "Attempting to unsubscribe from the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_C2D_SUBSCRIBE_TOPIC ) );
```

source/azure_iot_hub_client.c
danewalton: should we wipe the receive contexts on a connect?

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+    }
+    else
+    {
+        pxContext = &pxAzureIoTHubClient->_internal.xReceiveContext[ azureiothubRECEIVE_CONTEXT_INDEX_METHODS ];
+        xMqttSubscription.xQoS = eAzureIoTMQTTQoS0;
+        xMqttSubscription.pcTopicFilter = ( const uint8_t * ) AZ_IOT_HUB_CLIENT_METHODS_SUBSCRIBE_TOPIC;
+        xMqttSubscription.usTopicFilterLength = ( uint16_t ) sizeof( AZ_IOT_HUB_CLIENT_METHODS_SUBSCRIBE_TOPIC ) - 1;
+        usSubscribePacketIdentifier = AzureIoTMQTT_GetPacketId( &( pxAzureIoTHubClient->_internal.xMQTTContext ) );
+
+        AZLogDebug( ( "Attemptting to subscribe to the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_METHODS_SUBSCRIBE_TOPIC ) );
```
danewalton, line 894:
```suggestion
        AZLogDebug( ( "Attempting to subscribe to the MQTT topic: %s.", AZ_IOT_HUB_CLIENT_METHODS_SUBSCRIBE_TOPIC ) );
```

source/azure_iot_hub_client.c
```diff
@@ -626,3 +740,527 @@ AzureIoTHubClientResult_t AzureIoTHubClient_SendTelemetry( AzureIoTHubClient_t *
...
+{
+    AzureIoTMQTTResult_t xMQTTResult;
+    AzureIoTHubClientResult_t xReturn;
+    AzureIoTMQTTPublishInfo_t xMQTTPublishInfo = { 0 };
+    uint8_t ucRequestID[ azureiothubMAX_SIZE_FOR_UINT32 ];
+    size_t xTopicLength;
+    az_result xCoreResult;
+    az_span xRequestID = az_span_create( ucRequestID, sizeof( ucRequestID ) );
+
+    if( ( pxAzureIoTHubClient == NULL ) ||
```
danewalton, line 1153: should we be checking `pulRequestID` here?

hihigupt: it is optional

---
In reply to: [612835601](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612835601) [](ancestors = 612835601)

source/azure_iot_hub_client.c
hihigupt: no, in cases when user re-connect with clean session false, he should be allowed to receive messages without subscribe call.

---
In reply to: [612833270](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612833270) [](ancestors = 612833270)

source/azure_iot_hub_client.c
```diff
@@ -346,6 +356,110 @@ static uint32_t prvGetTimeMs( void )
...
+                                                      az_span xSpan,
+                                                      bool xOddSeq,
+                                                      uint32_t * pulRequestId,
+                                                      az_span * pxSpan )
+{
+    AzureIoTHubClientResult_t xReturn;
+    az_span xRemainder;
+    az_result xCoreResult;
+
+    /* Check the last requestId used, and do increment of 2 or 1 based on xOddSeq */
```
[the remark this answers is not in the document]
hihigupt: this is request Id.... not some variable.

---
In reply to: [612814933](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612814933) [](ancestors = 612814933)

source/azure_iot_hub_client.c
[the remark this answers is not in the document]
hihigupt: We do not expose span and all usage are internal.
I will keep it same for now.... until we think otherwise.

---
In reply to: [612817618](https://github.com/Azure/azure-iot-middleware-freertos/pull/75#discussion_r612817618) [](ancestors = 612817618)

source/azure_iot_hub_client.c
jspaith: @hihigupt - tagging because you'd resolved this discussion.  I literally woke up at 3AM last night and realized there's a problem with what I've proposed, too.  Suppose we set the callback pointers after `prvWaitForSubAck` and case where the PUBLISH comes right away, before we return.  I don't think this is outside the realm of possibility with say a C2D message that the Hub had queued to send.

In this case, if we don't have callback filled, then we end up with the case where the PUBLISH to us is dropped.  Which seems worse to me than case where we do the app's callback from within the subscribe function.

I don't have much appetite for complicated state machines where we mark the message wants to be PUBLISH'd to app, but don't send it to app here.  Beyond complexity I don't know how to do that without making a copy of the message.

So I think that we should just document clearly that there are cases where the callback may be called before we return from this function and that the application should be prepared for this case.

source/azure_iot_hub_client.c
jspaith: OK - the re-subscribe was only scenario that worried me here, so unless/until we can think of other issues agree with keeping this simple.

jspaith (approved):
Looks good, though my feedback yesterday about setting the callback structures after SUBSCRIBE I think was bad and we'll need to fix to be more like it was or else we can talk.  I've put more inline.

source/azure_iot_hub_client.c
hihigupt: Great point ... Setting up of receive path before we do Process Loop is important as we might have windows to drop packets. I will set the receive path and cleanup if failed

danewalton (approved):

PR #76, opened 2021-04-08
Add unit test for message api

hihigupt:


tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+
+    assert_int_equal( AzureIoT_MessagePropertiesInit( &xTestMessageProperties,
+                                                      ucBuffer, 0, sizeof( ucBuffer ) ),
+                      eAzureIoTSuccess );
+
+    assert_int_equal( AzureIoT_MessagePropertiesAppend( &xTestMessageProperties,
+                                                        ucTestKey, sizeof( ucTestKey ) - 1,
+                                                        ucTestValue, sizeof( ucTestValue ) - 1 ),
+                      eAzureIoTSuccess );   
+}
```
danewalton, line 98: note we should probably have a test in the future where we initialize a property bag with properties alread written and then append to it.

danewalton (approved):

source/azure_iot.c
```diff
@@ -308,7 +308,8 @@ AzureIoTResult_t AzureIoT_MessagePropertiesInit( AzureIoTMessageProperties_t * p
     az_span xPropertyBufferSpan = az_span_create( pucBuffer, ( int32_t ) ulBufferLength );
     az_result xResult;
 
-    if( pxMessageProperties == NULL )
+    if( ( pxMessageProperties == NULL ) ||
+        ( pucBuffer == NULL ) )
```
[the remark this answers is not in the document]
hihigupt: Separate PR that take care of all the places that uses Embedded sdk.

tests/azure_iot_ut.c
hihigupt: I am not sure which path will that check? Is that test already part of embedded sdk?

tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTMessagePropertiesAppend_Failure( void ** ppvState )
+{
+    AzureIoTMessageProperties_t xTestMessageProperties;
+
+    ( void ) ppvState;
+
+    assert_int_equal( AzureIoT_MessagePropertiesInit( &xTestMessageProperties,
```
jspaith, line 60: Nit: Missing comment about what this is testing.  I love that in the other places by the way...

tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+                                                      ucBuffer, 0, sizeof( ucTestKey ) ),
+                      eAzureIoTSuccess );
+
+    /* Failed for NULL key passed */
+    assert_int_not_equal( AzureIoT_MessagePropertiesAppend( &xTestMessageProperties,
```
jspaith, line 65: Nit: Could we check for the actual expected error?  This would help pin behavior of what we return in various cases and if this changes, the UT would trip over it and we could have a discussion RE backcompat concerns if any.

tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+                                                            sizeof( ucTestValue ) - 1 ),
+                          eAzureIoTSuccess ); 
+
+    /* Failed for NULL value passed */
+    assert_int_not_equal( AzureIoT_MessagePropertiesAppend( &xTestMessageProperties,
+                                                            NULL, 0, ucTestValue,
+                                                            sizeof( ucTestValue ) - 1 ),
+                          eAzureIoTSuccess ); 
+
+    /* Failed for bigger data append */
```
jspaith, line 76: Nit - this almost threw me because it relies on how we set this up initially.  Maybe something like:

/* Failed for bigger data append - buffer was initialized to be size sizeof( ucTestKey ) and isn't big enough for request */

tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+
+static void testAzureIoTMessagePropertiesFind_Failure( void ** ppvState )
+{
+    AzureIoTMessageProperties_t xTestMessageProperties;
+    const uint8_t * pucOutValue;
+    uint32_t ulOutValueLength;
+
+    ( void ) ppvState;
+
+    assert_int_equal( AzureIoT_MessagePropertiesInit( &xTestMessageProperties,
```
jspaith, line 109: Same feedback here & line below about having the /* ... */ explaining test, even if obvious.  Comments any other places this may be missing.

tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+                                                      ucBuffer, 0, sizeof( ucBuffer ) ),
+                      eAzureIoTSuccess );
+
+    assert_int_equal( AzureIoT_MessagePropertiesAppend( &xTestMessageProperties,
```
jspaith, line 143: Even though it's technically re-testing the embedded SDK and I don't want to over do that, any chance you could do a few more append() here and search for those?  I typically find 3 to be a reasonable number on a passthrough like this.

tests/azure_iot_ut.c
```diff
@@ -0,0 +1,172 @@
...
+    assert_int_equal( ulOutValueLength, sizeof( ucTestValue ) - 1 );
+    assert_memory_equal( pucOutValue, ucTestValue, ulOutValueLength );
+}
+/*-----------------------------------------------------------*/
+
+uint32_t ulGetAllTests()
+{
+    const struct CMUnitTest tests[] =
+    {
+        cmocka_unit_test( testAzureIoTMessagePropertiesInit_Failure ),
```
jspaith, line 162: For completeness I think we should have a `AzureIoT_Init` and `AzureIoT_Deinit` UT here, even if it's just a simple call.  This will help with any code-coverage based on UT at minimum :), and good to have extra guard rails up.

jspaith (approved):

tests/azure_iot_ut.c
hihigupt: this one is setup for test appends in next 3 call.

PR #79, opened 2021-04-08
Fix transport impl

hihigupt:


demo/Common/mimxrt1060/transport/using_mbedtls/using_mbedtls.c
```diff
@@ -801,39 +798,49 @@ int32_t TLS_FreeRTOS_send( NetworkContext_t * pNetworkContext,
...
+        {
+            /* Sent data, so update the tally and keep looping. */
+            xWritten += ( size_t ) lResult;
+        }
+        else if( ( lResult == 0 ) || ( lResult == -pdFREERTOS_ERRNO_ENOSPC ) )
+        {
+            /* No data sent. The secure sockets
+            * API supports non-blocking send, so stop the loop but don't
+            * flag an error. */
+        	lResult = 0;
```
jspaith, line 824: 1 - I'm concerned on this - we break out of loop but we return to caller that we've written the `xWritten` which may not be accurate.  Is the socket guaranteed to write *all* the data or can it chose to just write some?

2 - Another question is on buffer lifetime management for caller?  After we return from `TLS_FreeRTOS_send` will the caller assume it doesn't need to keep buffer around anymore?  Will a non-blocking I/O make a copy of that buffer so this is OK, or is there risk of freeing?

3 - Also the `-pdFREERTOS_ERRNO_ENOSPC` - putting a negative in front of an error code - isn't a convention I've ever seen.  Please add a comment about what's up with this.

4 - My ;tl-dr here maybe should be "Can we (or should we) flip on blocking I/O and what happens if we do other than potentially some perf hit?"

demo/Common/mimxrt1060/transport/using_mbedtls/using_mbedtls.c
```diff
@@ -759,40 +759,37 @@ int32_t TLS_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
...
-        /* Mark these set of errors as a timeout. The libraries may retry read
-         * on these errors. */
-        tlsStatus = 0;
-    }
-    else if( tlsStatus < 0 )
-    {
-        LogError( ( "Failed to read data: mbedTLSError= %s : %s.",
-                    mbedtlsHighLevelCodeOrDefault( tlsStatus ),
-                    mbedtlsLowLevelCodeOrDefault( tlsStatus ) ) );
+        lResult = ( int ) xRead;
```
danewalton, line 783: ?
```suggestion
        lResult = ( int32_t ) xRead;
```

demo/Common/stm32h745/transport/using_mbedtls/using_mbedtls.c
```diff
@@ -759,40 +759,37 @@ int32_t TLS_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
...
-        /* Mark these set of errors as a timeout. The libraries may retry read
-         * on these errors. */
-        tlsStatus = 0;
-    }
-    else if( tlsStatus < 0 )
-    {
-        LogError( ( "Failed to read data: mbedTLSError= %s : %s.",
-                    mbedtlsHighLevelCodeOrDefault( tlsStatus ),
-                    mbedtlsLowLevelCodeOrDefault( tlsStatus ) ) );
+        lResult = ( int ) xRead;
```
danewalton, line 783:
```suggestion
        lResult = ( int32_t ) xRead;
```

demo/Common/stm32l4_l4+/transport/using_mbedtls/using_mbedtls.c
```diff
@@ -759,40 +759,37 @@ int32_t TLS_FreeRTOS_recv( NetworkContext_t * pNetworkContext,
...
-        /* Mark these set of errors as a timeout. The libraries may retry read
-         * on these errors. */
-        tlsStatus = 0;
-    }
-    else if( tlsStatus < 0 )
-    {
-        LogError( ( "Failed to read data: mbedTLSError= %s : %s.",
-                    mbedtlsHighLevelCodeOrDefault( tlsStatus ),
-                    mbedtlsLowLevelCodeOrDefault( tlsStatus ) ) );
+        lResult = ( int ) xRead;
```
danewalton, line 783:
```suggestion
        lResult = ( int32_t ) xRead;
```

danewalton (approved):

demo/Common/stm32h745/transport/using_mbedtls/using_mbedtls.c
```diff
@@ -801,39 +798,49 @@ int32_t TLS_FreeRTOS_send( NetworkContext_t * pNetworkContext,
...
+        lResult = mbedtls_ssl_write( &( pTlsTransportParams->sslContext.context ),
+                                     pBuffer + xWritten,
+                                     bytesToSend - xWritten );
+
+        if( lResult > 0 )
+        {
+            /* Sent data, so update the tally and keep looping. */
+            xWritten += ( size_t ) lResult;
+        }
+        else if( ( lResult == 0 ) || ( lResult == -pdFREERTOS_ERRNO_ENOSPC ) )
```
jspaith, line 819: Same feedback about RE `pdFREERTOS_ERRNO_ENOSPC ` from the function above.

jspaith (changes requested):
Main concerns are around the non-blocking i/o handling; it may be cool as-is but I don't fully understand from quick read of mbed tls docs around this.

PR #84, opened 2021-04-13
Add separate version for middleware

danewalton:
closes #84

### DIFF
--- a/source/azure_iot_hub_client.c
+++ b/source/azure_iot_hub_client.c
@@ -12,6 +12,7 @@
 #include "FreeRTOS.h"
 #include "task.h"
 
+#include "azure_iot_version.h"
 #include "azure_iot_mqtt.h"
 #include "azure/az_iot.h"
 #include "azure/core/az_version.h"
@@ -29,7 +30,7 @@
 #endif /* azureiothubSUBACK_WAIT_INTERVAL_MS */
 
 #ifndef azureiothubUSER_AGENT
-    #define azureiothubUSER_AGENT    "DeviceClientType=c%2F" AZ_SDK_VERSION_STRING "%28FreeRTOS%29"
+    #define azureiothubUSER_AGENT    "DeviceClientType=c%2F" azureiotVERSION_STRING "%28FreeRTOS%29"
 #endif /* azureiothubUSER_AGENT */
 
 /*
--- a/source/include/azure_iot_version.h
+++ b/source/include/azure_iot_version.h
@@ -15,7 +15,8 @@
 #ifndef AZURE_IOT_VERSION_H
 #define AZURE_IOT_VERSION_H
 
-#define STRINGIFY(x) #x
+#define _azureiotSTRINGIFY2( x )    # x
+#define _azureiotSTRINGIFY( x )     _azureiotSTRINGIFY2( x )
 
 /* Major numeric identifier. */
 #define azureiotVERSION_MAJOR         0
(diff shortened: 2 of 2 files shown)


source/azure_iot_hub_client.c
```diff
@@ -29,7 +30,7 @@
 #endif /* azureiothubSUBACK_WAIT_INTERVAL_MS */
 
 #ifndef azureiothubUSER_AGENT
-    #define azureiothubUSER_AGENT    "DeviceClientType=c%2F" AZ_SDK_VERSION_STRING "%28FreeRTOS%29"
+    #define azureiothubUSER_AGENT    "DeviceClientType=c%2F" azureiotVERSION_STRING "%28FreeRTOS%29"
```
hihigupt, line 33: Should we add Embedded SDK version into this string?

source/include/azure_iot_version.h
```diff
@@ -0,0 +1,34 @@
...
+ * are part of Azure SDK's internal implementation; we do not document these symbols
+ * and they are subject to change in future versions of the SDK which would break your code.
+ */
+
+#ifndef AZURE_IOT_VERSION_H
+#define AZURE_IOT_VERSION_H
+
+/*/ The version in string format used for telemetry following the `semver.org` standard */
+/*/ (https://semver.org). */
+#define azureiotVERSION_STRING        "0.0.1-alpha"
```
jspaith, line 20: It'd be safer if we could derive `azureiotVERSION_STRING` from the `azureiotVERSION_MAJOR` and friends below.  This way there's only one source of truth instead of possibility for this to be split up.

source/include/azure_iot_version.h
```diff
@@ -0,0 +1,34 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
+/* SPDX-License-Identifier: MIT */
+
+/**
+ * @file azure_iot_version.h
+ *
+ * @brief Provides version information.
```
jspaith, line 7: Nit - "version information for Azure IoT FreeRTOS middleware" or whatever the right terminology we need to use here.

jspaith (approved):

source/azure_iot_hub_client.c
jspaith: I *think* we can get away without it so we should see if we can to save some bandwidth.  @wduraes - would we do anything if we had embedded SDK version FreeRTOS was using?

The FreeRTOS layer we need but we should know the embedded SDK version with some level of accuracy based on FreeRTOS (we could obviosuly submodule update embedded SDK without a version bump of middleware, so we couldn't know exactly, but I think that's OK).

danewalton: Yea I really don't think it would be that hard to derive. And it saves the bandwidth. Whenever we do a release, if we name them right, we can always go back to them and check the embedded version if we really need it.

source/include/azure_iot_version.h
```diff
@@ -0,0 +1,34 @@
...
+/*/ Major numeric identifier. */
+#define azureiotVERSION_MAJOR         0
+
+/*/ Minor numeric identifier. */
+#define azureiotVERSION_MINOR         0
+
+/*/ Patch numeric identifier. */
+#define azureiotVERSION_PATCH         1
+
+/*/ Optional pre-release identifier. SDK is in a pre-release state when present. */
```
hihigupt, line 31: curious why comment starts with /*/ rather then /*?

danewalton: I think this was uncrustify. I can update it

source/include/azure_iot_version.h
danewalton: you mind giving the newest version another look? Just to make sure the macros are acceptable. This is how it renders:
image

jspaith: Yup the latest around stringify stuff looks good.

hihigupt (approved):

PR #87, opened 2021-04-14
[IoTHubClient] Add symmetric key set api

hihigupt:


source/include/azure_iot_hub_client.h
```diff
@@ -45,6 +45,8 @@ typedef enum AzureIoTHubClientResult
     eAzureIoTHubClientSubackWaitTimeout,  /**< There was timeout while waiting for SUBACK. */
     eAzureIoTHubClientTopicNotSubscribed, /**< Topic not subscribed. */
     eAzureIoTHubClientPublishFailed,      /**< Failed to publish. */
+    eAzureIoTHubClientSubscribeFailed,    /**< Failed to subscribe. */
+    eAzureIoTHubClientUnSubscribeFailed,  /**< Failed to unsubscribe. */
```
danewalton, line 49: nit: 
```suggestion
    eAzureIoTHubClientUnsubscribeFailed,  /**< Failed to unsubscribe. */
```

source/azure_iot_hub_client.c
```diff
@@ -583,6 +650,33 @@ void AzureIoTHubClient_Deinit( AzureIoTHubClient_t * pxAzureIoTHubClient )
 }
 /*-----------------------------------------------------------*/
 
+AzureIoTHubClientResult_t AzureIoTHubClient_SetSymmetricKey( AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                             const uint8_t * pucSymmetricKey,
+                                                             uint32_t ulSymmetricKeyLength,
+                                                             AzureIoTGetHMACFunc_t xHmacFunction )
```
danewalton, line 656: just to unify
```suggestion
                                                             AzureIoTGetHMACFunc_t xHMACFunction )
```

source/azure_iot_hub_client.c
```diff
@@ -460,6 +461,72 @@ static AzureIoTHubClientResult_t prvWaitForSubAck( AzureIoTHubClient_t * pxAzure
...
+    if( ulBufferLeft < ulHMACBufferLength )
+    {
+        AZLogError( ( "AzureIoTHubClient token generation failed with insufficient buffer size." ) );
+        return eAzureIoTHubClientOutOfMemory;
+    }
+
+    ulBufferLeft -= ulHMACBufferLength;
+    pucHMACBuffer = pucSASBuffer + ulSasBufferLen - ulHMACBufferLength;
+
+    if( AzureIoT_Base64HMACCalculate( pxAzureIoTHubClient->_internal.xHMACFunction,
```
danewalton, line 505: `if () != eAzureSuccess`

source/azure_iot_hub_client.c
```diff
@@ -460,6 +461,72 @@ static AzureIoTHubClientResult_t prvWaitForSubAck( AzureIoTHubClient_t * pxAzure
...
+                                      ucKey, ulKeyLen, pucSASBuffer, ulBytesUsed, pucSASBuffer + ulBytesUsed, ulBufferLeft,
+                                      pucHMACBuffer, ulHMACBufferLength, &ulBytesUsed ) )
+    {
+        AZLogError( ( "AzureIoTHubClient failed to encoded HMAC hash" ) );
```
danewalton, line 509:
```suggestion
        AZLogError( ( "AzureIoTHubClient failed to encode HMAC hash" ) );
```

source/azure_iot_hub_client.c
```diff
@@ -583,6 +650,33 @@ void AzureIoTHubClient_Deinit( AzureIoTHubClient_t * pxAzureIoTHubClient )
...
+                                                             uint32_t ulSymmetricKeyLength,
+                                                             AzureIoTGetHMACFunc_t xHmacFunction )
+{
+    AzureIoTHubClientResult_t xResult;
+
+    if( ( pxAzureIoTHubClient == NULL ) ||
+        ( pucSymmetricKey == NULL ) || ( ulSymmetricKeyLength == 0 ) ||
+        ( xHmacFunction == NULL ) )
+    {
+        AZLogError( ( "AzureIoTHubClient set symmetric key failed: Invalid argument" ) );
```
danewalton, line 664: should this just be the function name?
```suggestion
        AZLogError( ( "AzureIoTHubClient_SetSymmetricKey failed: Invalid argument" ) );
```

source/azure_iot_hub_client.c
```diff
@@ -583,6 +650,33 @@ void AzureIoTHubClient_Deinit( AzureIoTHubClient_t * pxAzureIoTHubClient )
...
+        xResult = eAzureIoTHubClientInvalidArgument;
+    }
+    else
+    {
+        pxAzureIoTHubClient->_internal.pucSymmetricKey = pucSymmetricKey;
+        pxAzureIoTHubClient->_internal.ulSymmetricKeyLength = ulSymmetricKeyLength;
+        pxAzureIoTHubClient->_internal.pxTokenRefresh = prvIoTHubClientGetToken;
```
danewalton, line 671: might be a dumb question but is there ever a scenario where someone wouldn't want the refresh done automatically?

danewalton (approved):

source/azure_iot_hub_client.c
hihigupt: cert case, where you do not need SAS

source/azure_iot_hub_client.c
```diff
@@ -460,6 +461,72 @@ static AzureIoTHubClientResult_t prvWaitForSubAck( AzureIoTHubClient_t * pxAzure
...
+    ulBufferLeft = ulSasBufferLen - ulBytesUsed;
+
+    if( ulBufferLeft < ulHMACBufferLength )
+    {
+        AZLogError( ( "AzureIoTHubClient token generation failed with insufficient buffer size." ) );
+        return eAzureIoTHubClientOutOfMemory;
+    }
+
+    ulBufferLeft -= ulHMACBufferLength;
+    pucHMACBuffer = pucSASBuffer + ulSasBufferLen - ulHMACBufferLength;
```
jspaith, line 503: I think a comment here would help.  Basically you're using the last 48 bytes of the app's passed buffer to store the HMAC, the first part of the buffer is the result of the `az_iot_hub_client_sas_get_signature` call, and there's a middle section of buffer between end of `az_iot_hub_client_sas_get_signature` output and before the reserved 48 bytes that you need for AzureIoT_Base64HMACCalculate`s `pucBuffer` parameter, correct?

This all looks good, but it took me a bit to figure out how it all fits together.  So a comment about "only want caller to have to pass one buffer so we use different portions of the space as follows..." by time we're at HMAC time would've made this easier.

jspaith (approved):

PR #88, opened 2021-04-15
[IoTHubClient] Add unit tests

hihigupt:


tests/azure_iot_hub_client_ut.c
```diff
@@ -381,6 +474,690 @@ static void testAzureIoTHubClient_SendTelemetry_Success( void ** state )
...
+    will_return( AzureIoTMQTT_ProcessLoop, eAzureIoTMQTTSuccess );
+    xPacketInfo.ucType = azureiotmqttPACKET_TYPE_SUBACK;
+    xDeserializedInfo.usPacketIdentifier = usTestPacketId;
+    assert_int_equal( AzureIoTHubClient_SubscribeDeviceTwin( &xTestIoTHubClient,
+                                                             prvTestDeviceTwin,
+                                                             NULL, ( uint32_t ) -1 ),
+                      eAzureIoTHubClientSuccess );
+
+    will_return( AzureIoTMQTT_Publish, eAzureIoTMQTTSuccess );
+    pucPublishPayload = NULL;
```
danewalton, line 1102: what are these used for? It's not immediately obvious

tests/azure_iot_hub_client_ut.c
```diff
@@ -381,6 +474,690 @@ static void testAzureIoTHubClient_SendTelemetry_Success( void ** state )
...
+
+    will_return( AzureIoTMQTT_Publish, eAzureIoTMQTTSendFailed );
+    assert_int_equal( AzureIoTHubClient_SendMethodResponse( &xTestIoTHubClient,
+                                                            &req, 200, ucTestMethodResponsePayload,
+                                                            sizeof( ucTestMethodResponsePayload ) ),
+                      eAzureIoTHubClientPublishFailed );
+}
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTHubClient_SendMethodResponse_EmptyResponseSuccess( void ** state )
```
danewalton, line 895: do we care about checking if methods are subscribed to before sending a response?

danewalton (approved):

tests/azure_iot_hub_client_ut.c
hihigupt: Test MQTT port has check ... if the payload generated via middleware is equal to the test expected payload. Thich check happens only if test has set the expected MQTT payload.

hihigupt: https://github.com/Azure/azure-iot-middleware-freertos/blob/33db1d22451dec6b07302de55c9ba70c711f0f30/tests/azure_iot_cmocka_mqtt.c#L94

tests/azure_iot_hub_client_ut.c
hihigupt: As method response is fire and forget, do we care? We are not waiting for any response.

danewalton: I mean if someone tries to send a response for some weird reason without having first received a method call, is that something we care about? Or do we let it go through and maybe disconnect from the service (if that's what hub would do if it received a response for a method it never invoked).

tests/azure_iot_hub_client_ut.c
```diff
@@ -381,6 +475,690 @@ static void testAzureIoTHubClient_SendTelemetry_Success( void ** state )
...
+    AzureIoTHubClient_t xTestIoTHubClient;
+
+    ( void ) state;
+
+    prvSetupTestIoTHubClient( &xTestIoTHubClient );
+
+    will_return( AzureIoTMQTT_Subscribe, eAzureIoTMQTTSuccess );
+    will_return( AzureIoTMQTT_ProcessLoop, eAzureIoTMQTTSuccess );
+    xPacketInfo.ucType = azureiotmqttPACKET_TYPE_SUBACK;
+    xDeserializedInfo.usPacketIdentifier = usTestPacketId;
```
jspaith, line 584: Not directly related to this line, but code in question isn't in PR.  For the cmocka mock `AzureIoTMQTT_Init` and others, there's a global called `xTestUserCallback`.  This really confused me when trying to follow the path, because I assumed it meant user as far as getting to application layer and I couldn't see how it fit into the mock.  After more review I saw it's really referring to the middleware layer's MqttEvent processing callback.

I don't know how to better wordsmith that - maybe like `xMiddlewareCallback` - but if the cmocka layer could use that in function params and globals I think code would be easier to follow.

tests/azure_iot_hub_client_ut.c
```diff
@@ -381,6 +475,690 @@ static void testAzureIoTHubClient_SendTelemetry_Success( void ** state )
...
+    will_return( AzureIoTMQTT_Subscribe, eAzureIoTMQTTSuccess );
+    will_return( AzureIoTMQTT_ProcessLoop, eAzureIoTMQTTRecvFailed );
+    assert_int_equal( AzureIoTHubClient_SubscribeCloudToDeviceMessage( &xTestIoTHubClient,
+                                                                       prvTestCloudMessage,
+                                                                       NULL, ( uint32_t ) -1 ),
+                      eAzureIoTHubClientFailed );
+}
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTHubClient_CloudMessageSubscribe_Success( void ** state )
```
jspaith, line 573: Could we add a new test where AzureIoTHubClient_SubscribeCloudToDeviceMessage (and analogous other subscribes) are called multiple times, successfully.  It's possible an app could do this - like getting confused on state - and since MQTT protocol and our SDK allow this, we should make it works.

tests/azure_iot_hub_client_ut.c
```diff
@@ -381,6 +475,690 @@ static void testAzureIoTHubClient_SendTelemetry_Success( void ** state )
...
+
+    prvSetupTestIoTHubClient( &xTestIoTHubClient );
+
+    will_return( AzureIoTMQTT_Unsubscribe, eAzureIoTMQTTSendFailed );
+    assert_int_equal( AzureIoTHubClient_UnsubscribeCloudToDeviceMessage( &xTestIoTHubClient ),
+                      eAzureIoTHubClientUnsubscribeFailed );
+}
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTHubClient_CloudMessageUnsubscribe_Success( void ** state )
```
jspaith, line 758: Nit - should this be `testAzureIoTHubClient_UnsubscribeCloudToDeviceMessage` so there's symmetry in test naming and API tested?

jspaith, line 758: I think we should have a test like this, where unsubscribe is called without much setup.  And another test where we go through a successful subscription and then unsubscribe.

I get the way that the state is maintained now it doesn't matter, but this could change in the future and having the test as a backstop like this more faithfully mimics normal app scenario.

tests/azure_iot_hub_client_ut.c
```diff
@@ -381,6 +475,690 @@ static void testAzureIoTHubClient_SendTelemetry_Success( void ** state )
...
+                      eAzureIoTHubClientSuccess );
+
+    will_return( AzureIoTMQTT_Publish, eAzureIoTMQTTSuccess );
+    pucPublishPayload = NULL;
+    assert_int_equal( AzureIoTHubClient_GetDeviceTwin( &xTestIoTHubClient ),
+                      eAzureIoTHubClientSuccess );
+}
+/*-----------------------------------------------------------*/
+
+static void testAzureIoTHubClient_ReceiveMessages_Success( void ** state )
```
jspaith, line 1109: Do we have a test where we get a "rando" topic we're not subscribed to?  If not, I think we should add one.  This shouldn't happen, but I'd like to test that (a) no callbacks are called and (b) we don't crash/hang.

jspaith (approved):
There's a mix of (mostly) nits here, though there's some additional test cases I think we need around this as well.  I'm approving since I know you can tell the difference:).  If you want me to look at anything again though or have pushback on new tests just let me know of course. 

Thanks

tests/azure_iot_hub_client_ut.c
```diff
@@ -26,14 +47,39 @@ extern const uint8_t * pucPublishPayload;
...
 };
-
+static uint32_t ulReceivedCallbackFunctionId;
+static const ReceiveTestData_t xTestReceiveData[] =
+{
+    { .pucTopic = ( const uint8_t * ) testCLOUD_MESSAGE_TOPIC,
+      .ulTopicLength = sizeof( testCLOUD_MESSAGE_TOPIC ) - 1,
+      .pucPayload = ( const uint8_t * ) testCLOUD_MESSAGE,
+      .ulPayloadLength = sizeof( testCLOUD_MESSAGE ) - 1,
+      .ulCallbackFunctionId = testCLOUD_CALLBACK_ID },
```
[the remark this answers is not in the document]
hihigupt: Ack

tests/azure_iot_hub_client_ut.c
```diff
@@ -68,6 +114,39 @@ static void prvSetupTestIoTHubClient( AzureIoTHubClient_t * pxTestIoTHubClient )
 }
 /*-----------------------------------------------------------*/
 
+static void prvTestCloudMessage( AzureIoTHubClientCloudToDeviceMessageRequest_t * pxMessage,
```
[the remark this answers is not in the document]
hihigupt: sure

tests/azure_iot_hub_client_ut.c
hihigupt: Update the name to make it little better to understand the flow

hihigupt:
> There's a mix of (mostly) nits here, though there's some additional test cases I think we need around this as well. I'm approving since I know you can tell the difference:). If you want me to look at anything again though or have pushback on new tests just let me know of course.
> 
> Thanks

I have update the PR with new tests.

PR #89, opened 2021-04-15
Add comment about suback

### LINKED ISSUE #65: Inspect packet to see if ack was successful
`TODO` tag here:
https://github.com/Azure/azure-iot-middleware-freertos/blob/8de270669263019f735b59dfd1d7bcdf6a0521a3/source/azure_iot_hub_client.c#L90-L91

danewalton:
Closes #65

Hub will terminate if there is an error on their end processing the sub. If successful on their end, they will send a suback. It doesn't matter whether we receive it or not since if there was an error, the connection would be terminated.

### DIFF
--- a/source/azure_iot_hub_client.c
+++ b/source/azure_iot_hub_client.c
@@ -109,7 +109,7 @@ static void prvMQTTProcessSuback( AzureIoTHubClient_t * pxAzureIoTHubClient,
 
         if( pxContext->_internal.usMqttSubPacketID == usPacketId )
         {
-            /* TODO: inspect packet to see is ack was successful*/
+            /* We assume success since IoT Hub would disconnect if there was a problem subscribing. */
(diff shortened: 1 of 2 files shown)


source/azure_iot_hub_client.c
```diff
@@ -109,7 +109,7 @@ static void prvMQTTProcessSuback( AzureIoTHubClient_t * pxAzureIoTHubClient,
 
         if( pxContext->_internal.usMqttSubPacketID == usPacketId )
         {
-            /* TODO: inspect packet to see is ack was successful*/
+            /* We assume success since IoT Hub would disconnect if there was a problem subscribing. */
```
hihigupt, line 112: Can we also add same comment in provisioning client 

https://github.com/Azure/azure-iot-middleware-freertos/blob/33db1d22451dec6b07302de55c9ba70c711f0f30/source/azure_iot_provisioning_client.c#L566

danewalton: added

hihigupt (approved):

PR #96, opened 2021-04-21
Add CMake for sources and top level

danewalton:


CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+if(build_tests)
+  include(CTest)
+  enable_testing()
+
+  # Turn on the flags necessary for code coverage
+  add_compile_options(-fprofile-arcs -ftest-coverage -O0 -g3)
+  add_link_options(-fprofile-arcs -ftest-coverage)
+
+  # Set the port and config files
+  set(AZURE_IOT_MQTT_PORT ${CMAKE_CURRENT_LIST_DIR}/tests)
```
hihigupt, line 33: is there a convention to name variable (capital like marco or camel case).

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+# Add the middleware library
+add_subdirectory(source/)
+
+# Tests (mocks require GNU)
+if(build_tests AND ${CMAKE_COMPILER_IS_GNUCC})
+  set(MOCK_LINKER_OPTIONS "-Wl,--gc-sections")
+  add_subdirectory(tests/)
+elseif(build_tests AND NOT ${CMAKE_COMPILER_IS_GNUCC})
+  message(FATAL_ERROR "Due to necessary linker flags (for mocking functions), testing is only available on GNU gcc")
+endif()
```
hihigupt, line 56: nit:add newline

eng/cmake/global_compile_options.txt
```diff
@@ -0,0 +1,26 @@
+if(MSVC)
```
hihigupt, line 1: copywrite?

hihigupt: GCC?

source/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+add_library(az_iot_middleware_freertos
+  ${CMAKE_CURRENT_LIST_DIR}/azure_iot_hub_client.c
+  ${CMAKE_CURRENT_LIST_DIR}/azure_iot_provisioning_client.c
+  ${CMAKE_CURRENT_LIST_DIR}/azure_iot.c
+)
+
+# The user needs to supply this
+target_link_libraries(az_iot_middleware_freertos
+  PUBLIC
+    middleware_config)
```
hihigupt, line 18: this should be target_include_directories as config are headers?

hihigupt: with PRIVATE

source/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+
+target_link_libraries(az_iot_middleware_freertos
+  PUBLIC
```
hihigupt, line 21: private right?

source/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    az::iot::hub
+    az::iot::provisioning
+)
+
+if(WIN32)
+  target_include_directories(az_iot_middleware_freertos
+    PRIVATE
+      ${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS/Source/portable/MSVC-MingW
+  )
+  elseif(UNIX)
```
hihigupt, line 32: nit align

source/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+  # Include coreMQTT port
+  target_sources(az_iot_middleware_freertos
+    PRIVATE
+      ${az_iot_middleware_freertos_SOURCE_DIR}/port/coreMQTT/azure_iot_core_mqtt.c
+  )
+
+  target_include_directories(az_iot_middleware_freertos
+    PUBLIC
+      ${az_iot_middleware_freertos_SOURCE_DIR}/port/coreMQTT
+)
```
hihigupt, line 73: nit align

tests/CMakeLists.txt
```diff
@@ -1,123 +1,66 @@
...
-include_directories(${AZ_IOT_REPO_ROOT}/libraries/FreeRTOS/FreeRTOS/Source/include)
+include_directories("${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS/Source/include")
 
-# port for FreeRTOS
-if(UNIX)
-    include_directories(${AZ_IOT_REPO_ROOT}/libraries/FreeRTOS/FreeRTOS/Source/portable/ThirdParty/GCC/Posix)
-else()
-    message( FATAL_ERROR "Tests only supported in linux." )
+if(WIN32)
+  include_directories("${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS/Source/portable/MSVC-MingW")
```
hihigupt, line 10: this is control via source CMAKE right?

tests/CMakeLists.txt
```diff
@@ -1,123 +1,66 @@
...
+    ${DEFAULT_C_COMPILE_FLAGS}
+  LINK_LIBRARIES
+    cmocka
+    az::iot_middleware::freertos
+    middleware_config
+  LINK_OPTIONS ${MOCK_LINKER_OPTIONS}
+  INCLUDE_DIRECTORIES
+    ${CMOCKA_INCLUDE_DIR}
+    ${CMAKE_CURRENT_LIST_DIR}
+)
```
hihigupt, line 66: nit \n

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+cmake_minimum_required (VERSION 3.13)
```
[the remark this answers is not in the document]
danewalton: This is what amazon freertos uses so I'd say stay aligned with that since if someone uses that it'll have to be 3.13 anyways.

CMakeLists.txt
danewalton: Yup usually capitals snake case

source/CMakeLists.txt
danewalton: It's built as an interface library which means it's only headers.

CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
...
+
+project(az_iot_middleware_freertos LANGUAGES C)
+
+set(CMAKE_C_STANDARD 99)
+
+# Add modules and include compiler options/switches
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake-modules")
+# include(eng/cmake/global_compile_options.txt)
```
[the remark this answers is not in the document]
danewalton: Added to only add the options to our repo

.github/scripts/ci_tests.sh
```diff
@@ -11,20 +11,27 @@ TEST_RUN_E2E_TESTS=${1:-1}
 TEST_CORES=${2:-2}
 TEST_JOB_COUNT=${3:-2}
 
-echo -e "Building tests"
-cmake tests -Bbuild
+echo -e "::group::Building unit tests"
+cmake -Dbuild_ut_tests=ON -Bbuild .
```
hihigupt, line 15: >Bbuild [](http://example.com/codeflow?start=27&length=6)

lets keep it at the end .... because there are many logs emitted via e2e test and its hard to see the code coverage in middle.

eng/cmake/global_compile_options.cmake
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
hihigupt, line 1: Can we club these cmake stuff under cmake root dir?

source/CMakeLists.txt
hihigupt: will there be config.a or its all header?

source/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+    az::iot::hub
+    az::iot::provisioning
+)
+
+# Include port files
+# TODO: Have the user select the port path to use
+if(WIN32)
+  target_include_directories(az_iot_middleware_freertos
+    PRIVATE
+      ${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS/Source/portable/MSVC-MingW
```
hihigupt, line 32: Can we have port target of FreeRTOS and user define it ? Here we just consume it

source/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+elseif(UNIX)
+  target_include_directories(az_iot_middleware_freertos
+    PRIVATE
+      ${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS/Source/portable/ThirdParty/GCC/Posix
+  )
+endif()
+
+# Include FreeRTOS headers
+target_include_directories(az_iot_middleware_freertos
+  PUBLIC
```
hihigupt, line 43: >PUBLIC [](http://example.com/codeflow?start=2&length=6)

Possibly target for FreeRTOS? and here we depend upon that targett?

source/CMakeLists.txt
```diff
@@ -0,0 +1,90 @@
...
+
+target_include_directories(az_iot_middleware_freertos
+    PUBLIC
+      ${CMAKE_CURRENT_LIST_DIR}/include
+      ${CMAKE_CURRENT_LIST_DIR}/interface
+)
+
+# Check if custom mqtt port path is set, otherwise
+# use default coreMQTT port
+if(AZURE_IOT_MQTT_PORT)
```
hihigupt, line 55: >AZURE_IOT_MQTT_PORT [](http://example.com/codeflow?start=3&length=19)

Make me think that having three target this lib compilation depends upon.

1, FreeRTOS
2. FreeRTOS port.
3. MQTT port.

and these target can be define by application. What you think?

tests/e2e/CMakeLists.txt
```diff
@@ -0,0 +1,226 @@
+
+############################################# Add E2E Tests ################################################
```
hihigupt, line 2: ms banner? and min version check?

tests/e2e/CMakeLists.txt
```diff
@@ -0,0 +1,226 @@
...
+
+
+set(FREERTOS_ROOT_PATH "${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS")
+set(FREERTOS_PLUS_ROOT_PATH "${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS-Plus")
+
+include_directories(${CMAKE_CURRENT_LIST_DIR}/../config_files)
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -m32")
+add_compile_options(-m32 -ggdb3 -O0 -D_DEBUG -DprojCOVERAGE_TEST=0 -DMBEDTLS_CONFIG_FILE=\"mbedtls_config.h\" -D_WINDOWS_)
```
hihigupt, line 11: >D_WINDOWS_ [](http://example.com/codeflow?start=111&length=10)

Windows?

danewalton: these are all copied from the initial CMake that you had added. I wasn't sure if this was something you guys found was needed in Azure RTOS.

.github/scripts/ci_tests.sh
danewalton: image

I added the `::group::` tags in the logs which group each subsection under a collapsible menu. I `rm -rf build` the build directory between the ut and e2e since each needs its own build tree and wasn't sure if that messes with the code coverage.

eng/cmake/global_compile_options.cmake
danewalton: Re-organized the compile options and modules to both be under a `/cmake` directory. Now we have `/cmake/build` for compile and `/cmake/modules` which is a common practice for modules.

source/CMakeLists.txt
danewalton: This part requires a bit of thinking I feel. We could go the target route which requires a bit of tweaking on their end since they might not have it. They would have to tweak their own CMake build to construct something that WE need. Fundamentally we just need a path so maybe we ask them to set an environment variable kind of like the embedded for the `VCPKG_ROOT`? That way less of our build requirements leak into their build process?

tests/e2e/CMakeLists.txt
danewalton: Ah yup this is left over when e2e and ut were combined in one. For the min version check though that only needs to be before the first `project()` call which is in the top level. It defines the min version for the whole project there.

source/CMakeLists.txt
danewalton: I think I fundamentally agree with you. Something like three paths for those three parts? Environment variables?

.github/scripts/ci_tests.sh
```diff
@@ -11,20 +11,27 @@ TEST_RUN_E2E_TESTS=${1:-1}
...
-echo -e "Running unit tests"
+echo -e "::group::Running unit tests"
 ctest -j $TEST_JOB_COUNT -C "debug" --output-on-failure --schedule-random -T test
 
+echo -e "::group::Code coverage"
+gcovr -r $(pwd) -f ../source/.*.c
+
+popd
+
+echo -e "::group::Building e2e tests"
```
hihigupt, line 27: E2E test running it two step process .
1. Building device side binary: require some config setting up and then building with those config.
2. Compile service side scripts and run them.

These both steps are coded in the script (./tests/e2e/run.sh). You do not need to compile code here. Check this file
https://github.com/Azure/azure-iot-middleware-freertos/blob/cbb4585fc6a903ea90264556cd23bcede882e74f/tests/e2e/run.sh#L40

Possibly this special echo can be done in that file.

CMakeLists.txt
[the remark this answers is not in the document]
hihigupt: As linux default installation is 3.10 and if we can are not depending on feature that is only present in 3.13 and onward. It is better to lower this bar.

.github/scripts/ci_tests.sh
```diff
@@ -11,20 +11,27 @@ TEST_RUN_E2E_TESTS=${1:-1}
 TEST_CORES=${2:-2}
 TEST_JOB_COUNT=${3:-2}
 
-echo -e "Building tests"
-cmake tests -Bbuild
+echo -e "::group::Building unit tests"
```
hihigupt, line 14: I see that currently we are mixing the 32 and 64 bit arch
https://github.com/Azure/azure-iot-middleware-freertos/pull/96/checks?check_run_id=2449924893#step:5:652

Can we make our cmake agnostic to this and let ci_test.sh to only select 32 bit now? 
-m32 flag for gcc help you compile 32bit binaries.

tests/e2e/CMakeLists.txt
hihigupt: now that we are using cmake ...lets remove these 
-m32 -ggdb3 -O0 -D_DEBUG

and do a top level build -DCMAKE_BUILD_TYPE=Debug and COMPILE_FLAGS "-m32" LINK_FLAGS "-m32"

tests/ut/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+cmake_minimum_required(VERSION 3.13)
+
+# FreeRTOS Kernel include
+include_directories(${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS/Source/include)
+include_directories(${az_iot_middleware_freertos_SOURCE_DIR}/libraries/FreeRTOS/FreeRTOS-Plus/Source/Utilities/logging)
```
hihigupt, line 8: this should no more be required in UT tests

tests/e2e/CMakeLists.txt
danewalton: Yea the `-m32` compile flag is handled at the top. The debug option can be handled by the command line so removed that too. Check again to see if the remaining flags look alright. It built on my machine this way so we'll see how the CI likes it.

.github/scripts/ci_tests.sh
danewalton: is that not a `node` problem?

hihigupt: I should not be ... node exec the binary and binary that should be build should be of 32 bin .... looks like currently binary is getting build with x64. Not sure why, as in early run logs (current master) I am not seeing this type of loading error.

.github/scripts/ci_tests.sh
danewalton: ahh I see didn't realize the script was building too. I thought it was just running the executable.

.github/scripts/ci_tests.sh
danewalton: I don't think this is introduced with this PR. [Link to a previous one with the same one.](https://github.com/Azure/azure-iot-middleware-freertos/runs/2416518140?check_suite_focus=true#step:5:433) We can look into it but I don't think it's to be fixed here.

hihigupt: ooh then its bit more then how you are building it here . I can take a look into this.

CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+# Add modules and include compiler options/switches
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/modules")
+include(cmake/build/compile_options.cmake)
+
+# Options for building the lib
+option(build_ut_tests "Build the unit tests for the Azure IoT Middleware for FreeRTOS" OFF)
+option(build_e2e_tests "Build the e2e tests for the Azure IoT Middleware for FreeRTOS" OFF)
+set(config_directory CACHE STRING "The directory which has the FreeRTOSConfig.h and azure_iot_config.h")
+set(freertos_directory CACHE STRING "The directory which has the source code for FreeRTOS (as structured on their Github repo)")
+set(port_directory CACHE STRING "The directory which has the port layer for FreeRTOS")
```
hihigupt, line 19: freertos_kernel_port_dir

hihigupt: or just freertos_port_dir

CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+# If testing, turn on relevant switches and add configurations.
+if(build_ut_tests OR build_e2e_tests)
+  include(CTest)
+  enable_testing()
+
+  # Include port files for all targets
+  if(WIN32)
+    include_directories(${freertos_repo_SOURCE_DIR}/FreeRTOS/Source/portable/MSVC-MingW)
+  elseif(UNIX)
+    include_directories(${freertos_repo_SOURCE_DIR}/FreeRTOS/Source/portable/ThirdParty/GCC/Posix)
```
hihigupt, line 66: this should always be required irrespective of e2e test or unit test

CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+  endif()
+
+  # Include config files for all targets
+  include_directories(${CMAKE_CURRENT_LIST_DIR}/tests/config_files)
```
hihigupt, line 70: can test cmake decide this ?

CMakeLists.txt
danewalton: What if someone is building for a board?

hihigupt: We should keep the RTOS port same throughout the targets (UT/E2E and lib) and error our in case when building e2e/ut with other ports.

source/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+  )
+
+  target_include_directories(az_iot_middleware_freertos
+    PUBLIC
+      ${az_iot_middleware_freertos_SOURCE_DIR}/ports/coreMQTT
+  )
+
+  target_link_libraries(az_iot_middleware_freertos
+    PUBLIC
+      coremqtt
```
hihigupt, line 62: >coremqtt [](http://example.com/codeflow?start=6&length=8)

cursious .. will this target get build when using custom non-core-mqtt client?

source/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+cmake_minimum_required(VERSION 3.13)
+
+set(CMAKE_C_STANDARD 99)
+
+# Add coreMQTT Lib
+add_library(coremqtt
```
hihigupt, line 9: >coremqtt [](http://example.com/codeflow?start=12&length=8)

coreMQTT target has dependency on config folder, but we are not providing it ... Can you check if this build when only lib is build?

CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
hihigupt, line 1: should this file be under root dir?

cmake/build/compile_options.cmake
```diff
@@ -0,0 +1,31 @@
...
+
+    target_compile_options(${_TARGET} PRIVATE -Xclang -Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wdocumentation -Wdocumentation-unknown-command -fcomment-block-commands=retval -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wfloat-equal)
+  else()
+    if(WARNINGS_AS_ERRORS)
+      set(WARNINGS_AS_ERRORS_FLAG "-Werror")
+    endif()
+
+    target_compile_options(${_TARGET} PRIVATE -Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wlogical-op -Wfloat-equal)
+  endif()
+endfunction()
```
hihigupt, line 31: add \n

source/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+
+target_include_directories(az_iot_middleware_freertos
+    PUBLIC
+      ${CMAKE_CURRENT_LIST_DIR}/include
+      ${CMAKE_CURRENT_LIST_DIR}/interface
+)
+
+# Check if custom mqtt port path is set, otherwise
+# use default coreMQTT port
+if(NOT( "${AZURE_IOT_MQTT_PORT}" STREQUAL "" ))
```
hihigupt, line 49: Let keep variable in capital and target in small case with _

tests/e2e/CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -m32")
+
+include_directories(${CMAKE_CURRENT_LIST_DIR}/../config_files)# Include config
+
+add_compile_options(-DprojCOVERAGE_TEST=0 -DMBEDTLS_CONFIG_FILE=\"mbedtls_config.h\")
+
+message("Fetching FreeRTOS - This could take up to a minute or so...")
+include(FetchContent)
+FetchContent_Declare(
```
hihigupt, line 23: lets keep FreeRTOS dir empty so this fetching is done in one place i.e in root dir cmake

hihigupt: Want to make sure we do this pull only at one place

tests/ut/CMakeLists.txt
```diff
@@ -0,0 +1,92 @@
...
+# Turn on the flags necessary for code coverage
+add_compile_options(-fprofile-arcs -ftest-coverage -O0 -g3)
+add_link_options(-fprofile-arcs -ftest-coverage)
+set(MOCK_LINKER_OPTIONS "-Wl,--gc-sections")
+
+message("Fetching FreeRTOS - This could take up to a minute or so...")
+include(FetchContent)
+FetchContent_Declare(
+  freertos_repo
+  GIT_REPOSITORY        https://github.com/FreeRTOS/FreeRTOS.git
```
hihigupt, line 28: same

hihigupt (approved):

PR #97, opened 2021-04-21
[E2ETest] Add device side code

hihigupt:


tests/e2e/device/e2e_device_commands.c
```diff
@@ -0,0 +1,1150 @@
...
+ * Scan forward till key is not found
+ *
+ **/
+static uint32_t prvGetJsonValueForKey( az_json_reader * pxState,
+                                       const char * pcKey,
+                                       az_span * pxValue )
+{
+    uint32_t ulStatus = e2etestE2E_TEST_NOT_FOUND;
+    az_span xKeySpan = az_span_create( ( uint8_t * ) pcKey, strlen( pcKey ) );
+    int32_t ulLength;
```
hihigupt, line 208: lLenght

tests/e2e/device/Makefile
```diff
@@ -0,0 +1,206 @@
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
danewalton, line 2: would this file get replaced with CMake or leave it as Make?

danewalton (approved):

tests/e2e/device/Makefile
hihigupt: Yes, we will change it to cmake... once we have root cmake support.

tests/e2e/device/FreeRTOSConfig.h
```diff
@@ -0,0 +1,204 @@
+/*
```
jspaith, line 1: Nit: Could we put FreeRTOS Config type stuff into like a tests/common?  I'm worried that if we ever add new test cases, we'll end up with bunches of copies of this floating around when we really we should just need one?

If this ends up making stuff gross or we want to see till we get >1 test scenario that needs it, I'm fine if you punt this.

tests/e2e/device/config/azure_iot_config.h
```diff
@@ -0,0 +1,68 @@
...
+/* Map the SdkLog macro to the logging function to enable logging
+ * on Windows simulator. */
+#ifndef SdkLog
+    #define SdkLog( message )    vLoggingPrintf message
+#endif
+
+#include "logging_stack.h"
+/************ End of logging configuration ****************/
+
+#define azureiotROOT_CA_PEM "-----BEGIN CERTIFICATE-----\r\n" \
```
jspaith, line 46: Outside scope of this PR - so feel free to resolve this comment but we should discuss further - but what's our story around helping customers get the root CA right? 

If we have some `certs.c` equivalent from C SDK, it feels the header needs a lot more comments in it (or a link to official guidance).  And if we have such an equivalent setup, ideally the e2e could reference that as well.

tests/e2e/device/config/core_mqtt_config.h
```diff
@@ -0,0 +1,78 @@
+/*
```
jspaith, line 1: Similar to my question about freeRtosConfig.h above, could this / should this be in a tests/common folder??

tests/e2e/device/config/core_mqtt_config.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include "logging_levels.h"
+
+/* Logging configuration for the MQTT library. */
+#ifndef LIBRARY_LOG_NAME
+    #define LIBRARY_LOG_NAME    "MQTT"
+#endif
+
+#ifndef LIBRARY_LOG_LEVEL
+    #define LIBRARY_LOG_LEVEL    LOG_INFO
```
jspaith, line 47: Do we only want at LOG_INFO?  I get setting log level is hard because if we're too verbose we mess up timing, but I don't think we have *that* many logs.  Should this be LOG_DEBUG instead?  Imagine if we hit some weirdo gate failure -- in that case I'd think we'd want as much log output as possible. ??

tests/e2e/device/config/core_mqtt_config.h
```diff
@@ -0,0 +1,78 @@
...
+#endif
+
+/* Prototype for the function used to print to console on Windows simulator
+ * of FreeRTOS.
+ * The function prints to the console before the network is connected;
+ * then a UDP port after the network has connected. */
```
jspaith, line 53: Do we need to have a UDP port for e2e tests?  I feel that sticking with console is easier and is what we could get off of gates in any chance?

tests/e2e/device/e2e_device_commands.c
```diff
@@ -0,0 +1,1150 @@
+/* Copyright (c) Microsoft Corporation. All rights reserved. */
```
jspaith, line 1: I'm assuming (not so naively I believe) that this is refactored implementation from Azure RTOS, which itself has got extensive review?

If I'm wrong let me know and especially if there's areas you think may need more review, otherwise just resolve this comment.

jspaith: Analogous feedback for the other test implementations of .c files.

tests/e2e/device/config/azure_iot_config.h
```diff
@@ -0,0 +1,68 @@
...
+
+/* Prototype for the function used to print to console on Windows simulator
+ * of FreeRTOS.
+ * The function prints to the console before the network is connected;
+ * then a UDP port after the network has connected. */
+extern void vLoggingPrintf( const char * pcFormatString,
+                            ... );
+
+/* Map the SdkLog macro to the logging function to enable logging
+ * on Windows simulator. */
```
jspaith, line 38: Not this PR per se, but in the readme I see "Windows support will follow" after we only do Linux.

If not already, we should open an issue or work with Wellington to track fact we need Windows port if we're serious about "Windows support will follow" (which I think we should be given we have diversity of environments our customers are running on.)

tests/e2e/run.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+#
+# This script install dependencies of service side of e2e tests and runs e2e test
+
+# ./run.sh <RTOS connectivity interface>
+# e.g. ./run.sh veth1
+
+if [[ ! -z ${DEBUG_SHELL} ]]
```
jspaith, line 11: What is DEBUG_SHELL?  I looked both in our setup readme & general bash instructions and this isn't clear what this is?  

I feel we need a comment here I feel (unless my affinity for PowerShell is just making me totally clueless about this) and if there's special setup needed the readme.md should point user to more.

tests/e2e/run.sh
```diff
@@ -0,0 +1,43 @@
...
+}
+
+[ -n $1 ] || die "ethernet interface need to be passed; like veth1"
+
+filepath=`abspath $0`
+dir=`dirname $filepath`
+
+echo -e "Updating FreeRTOSConfig.h to point to interface $1"
+index=`tcpdump --list-interfaces | grep -Ei "([0-9]+).$1" | sed -E 's/^([0-9]+).*/\1/g'`
+sed -i "s/#define configNETWORK_INTERFACE_TO_USE.*/#define configNETWORK_INTERFACE_TO_USE ($index)/g" $dir/device/FreeRTOSConfig.h
```
jspaith, line 36: Is there a way we could not use `sed`.  Like any way under the sun:)?

Once we get cmake support, the lines below where you're doing `make` presumably that gets turned to a cmake 1st, right?  Could we use a cmake define "thingy" to enable this?

tests/e2e/run.sh
```diff
@@ -0,0 +1,43 @@
...
+
+echo -e "Building Device code"
+cd $dir/device; make
+
+echo -e "Running tests"
+export DEVICE_TEST_EXE="$dir/device/build/e2e_device_exe"
+cd $dir/service; ./mocha_exec.sh alltest
```
jspaith, line 43: Is there anyway that we could make it so that the `stdout|stderr|stdwrn` buffer size on this is 0.  I get there's some performance hit, and we should discuss if we think too much (I don't because network I/O takes up most of the clock cycles versus a write to console).

I'm asking this because I've seen enough cases on gate runs where we Log/Log/Log/CRASH!, but the interesting logs that would immediately help us diagnose the error are still in the stdout buffer that don't that log.

I looked around very briefly, maybe https://stackoverflow.com/questions/30546183/is-there-a-way-to-flush-stdout-of-a-running-process  ??  Though unlike top answers I don't think we want this output to a file since our gates are joined at hip to stdout & friends.

jspaith (approved):
If there's been extensive changes to the Azure RTOS version of the .c files here we should discuss offline, but otherwise looks good.  

It warms my heart so see e2e getting lit up this early btw:)!

hihigupt:
> If there's been extensive changes to the Azure RTOS version of the .c files here we should discuss offline, but otherwise looks good.
> 
> It warms my heart so see e2e getting lit up this early btw:)!

C files are mostly same, but differ in thread safety. This implementation make sure we are using single thread,

tests/e2e/device/config/azure_iot_config.h
hihigupt: Rotation of cert becomes complex, if we start distributing certs via our code. It is much better to let customer control it, we can help them to point to right cert to get via readme. Let me know what you think.

tests/e2e/device/config/core_mqtt_config.h
hihigupt: updated

tests/e2e/device/config/core_mqtt_config.h
hihigupt: ACK

tests/e2e/device/config/core_mqtt_config.h
hihigupt: updated. this was copy from coreMQTT

tests/e2e/run.sh
hihigupt: added comment to enable command trace

tests/e2e/run.sh
hihigupt: Index of interface is dynamic and will change based on which VM it is running so need to find index based on machine.

Yes make will be replaced via cmake.

tests/e2e/run.sh
hihigupt: https://linux.die.net/man/1/stdbuf

tests/e2e/device/FreeRTOSConfig.h
hihigupt: ack

tests/e2e/device/e2e_device_commands.c
hihigupt: this is similar to azrtos device side code .

tests/e2e/device/config/azure_iot_config.h
hihigupt: I will remove this comment ... but having build system for windows does not buys us anything other then testing it on one more simulated environment (compiled using different compiler).

tests/e2e/device/config/azure_iot_config.h
jspaith: Great point.  

Could we please put a comment in the test to this effect, like "/* This certificate is for test purposes only.  See official documentation about certificate management for your released product. */".  I don't think you actually need a link here, it's just enough to warn folks off from this one.

I sometimes go through test code to understand how stuff works and here I think we should call out this isn't the pattern - or at least there's more to understand here than copy/paste/forget.

PR #110, opened 2021-05-04
unify macro naming

danewalton:


source/include/azure_iot_config_defaults.h
```diff
@@ -60,8 +60,8 @@
  * @brief Max provisioning response payload supported.
  *
  */
-#ifndef azureiotPROVISIONING_REQUEST_PAYLOAD_MAX
-    #define azureiotPROVISIONING_REQUEST_PAYLOAD_MAX    ( 512U )
+#ifndef azureiotprovisioningREQUEST_PAYLOAD_MAX
```
hihigupt, line 63: >azureiotprovisioningREQUEST_PAYLOAD_MAX [](http://example.com/codeflow?start=8&length=39)

reason I name it like this :

<filename prefix<Capiltal macroname>>

In this case:
filename prefix: azureiot 
Macro: PROVISIONING_REQUEST_PAYLOAD_MAX

hihigupt (approved):

PR #112, opened 2021-05-05
add freertos uncrustify

danewalton:
Copy of the uncrustify config from FreeRTOS:
https://github.com/FreeRTOS/FreeRTOS/blob/main/tools/uncrustify.cfg


uncrustify.cfg
```diff
@@ -0,0 +1,160 @@
...
+mod_add_long_ifdef_endif_comment = 10      # unsigned number
+mod_add_long_ifdef_else_comment = 10       # unsigned number
+mod_case_brace                  = remove   # ignore/add/remove/force
+mod_remove_empty_return         = true     # false/true
+pp_indent                       = force    # ignore/add/remove/force
+pp_indent_at_level              = true     # false/true
+pp_indent_count                 = 4        # unsigned number
+pp_space                        = remove   # ignore/add/remove/force
+pp_if_indent_code               = true     # false/true
+# option(s) with 'not default' value: 158
```
hihigupt, line 160: \n

hihigupt (approved):

.github/scripts/code_style.sh
```diff
@@ -0,0 +1,54 @@
...
+    git checkout uncrustify-0.67
+    mkdir build
+    cd build
+    cmake ..
+    cmake --build .
+    sudo make install
+    cd ../..
+fi
+
+if [ $FIX -ne 0 ]; then
```
hihigupt, line 26: generally making explicit check is good thing... by doing -ne 0 increase the number of possible value hold true.

usage() {
    echo "${0} [check|fix]" 1>&2
    exit 1
}


FIX=${1:-"check"}


if [[ "$FIX" == "check" ]]; then


elif [[ "$FIX" == "check" ]] then

else
 usage();
fi

.github/scripts/code_style.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+set -o errexit # Exit if command failed.
+set -o nounset # Exit if variable not set.
+set -o pipefail # Exit if pipe failed.
+
+# Pass anything as a first parameter to fix code style problems
+FIX=${1:-0}
```
hihigupt, line 11: FIX=${1:-""}

hihigupt: remove this

.github/scripts/code_style.sh
```diff
@@ -0,0 +1,63 @@
...
+
+usage() {
+    echo "${0} [check|fix]" 1>&2
+    exit 1
+}
+
+FIX=${1:-"check"}
+
+# Version 0.67 is the source of truth
+if ! [ -x "$(command -v uncrustify)" ]; then
```
hihigupt, line 21: Can we create this in temp dir?

 tmp_dir=$(mktemp -d -t uncrustify-XXXX)
 pushd
 cd tmp_dir
 git clone https://github.com/uncrustify/uncrustify.git
 cd uncrustify
 git checkout uncrustify-0.67
 mkdir build
 cd build
 cmake ..
 cmake --build .
 sudo make install
 popd

.github/scripts/code_style.sh
danewalton: crap yup thanks for that

.github/scripts/code_style.sh
danewalton: oh didn't know that was a command. using

.github/scripts/code_style.sh
```diff
@@ -0,0 +1,62 @@
...
+    ./ports/coreMQTT/*.c                                      \
+    ./ports/coreMQTT/*.h                                      \
+    ./tests/config_files/*.h                                  \
+    ./tests/e2e/device/*.c                                    \
+    ./tests/e2e/device/*.h                                    \
+    ./tests/ut/*.h                                            \
+    ./tests/ut/*.c)
+
+    if [ $? -ne 0 ]; then
+      echo $RESULT | grep "FAIL"
```
hihigupt, line 46: I am assuming  uncrustify will return $? as non-zero value when failed. In that case lets exit with if we see that and fail the script.

exit $?

danewalton: The grep step attempts to parse through the large amount of text which it puts to the console to highlight what failed. I can remove it if necessary but I think it can be useful

hihigupt: Weird 
set -o errexit # Exit if command failed.

should never allow grep to be called.

hihigupt (approved):

PR #121, opened 2021-05-17
add build for clang

danewalton:
Example of the malformed doxygen docs being caught:
https://github.com/Azure/azure-iot-middleware-freertos/runs/2604875633#step:6:285


ports/coreMQTT/azure_iot_core_mqtt.c
```diff
@@ -190,7 +190,7 @@ uint16_t AzureIoTMQTT_GetPacketId( AzureIoTMQTTHandle_t xContext )
 
     xResult = MQTT_GetPacketId( xContext );
 
-    return prvTranslateToAzureIoTMQTTResult( xResult );
+    return ( uint16_t ) prvTranslateToAzureIoTMQTTResult( xResult );
```
hihigupt, line 193: >prvTranslateToAzureIoTMQTTResult [](http://example.com/codeflow?start=24&length=32)

:( its a bug. Can you do following thing:

uint16_t AzureIoTMQTT_GetPacketId( AzureIoTMQTTHandle_t xContext )
{
    return  MQTT_GetPacketId( xContext );
}

cmake/build/compile_options.cmake
```diff
@@ -20,7 +20,10 @@ function(az_add_compile_options _TARGET)
       set(WARNINGS_AS_ERRORS_FLAG "-Werror")
     endif()
 
-    target_compile_options(${_TARGET} PRIVATE -Xclang -Wall -Wextra -pedantic  ${WARNINGS_AS_ERRORS_FLAG} -Wdocumentation -Wdocumentation-unknown-command -fcomment-block-commands=retval -Wunused -Wuninitialized -Wmissing-declarations -Wconversion -Wpointer-arith -Wshadow -Wfloat-equal)
+    if(check_docs)
+      target_compile_options(${_TARGET} PRIVATE -Wdocumentation -Wdocumentation-unknown-command)
+    endif()
```
hihigupt, line 25: how about doing this check always.

.github/scripts/ci_tests.sh
```diff
@@ -25,6 +25,27 @@ fi
...
+set -e
+
+rm -rf build/
+
+echo -e "::group::Build using clang"
+cmake -Bbuild -DCMAKE_C_COMPILER=clang -Dfreertos_directory=$TEST_FREERTOS_SRC -Dfreertos_port_directory=$TEST_FREERTOS_SRC/FreeRTOS/Source/portable/ThirdParty/GCC/Posix -Dconfig_directory=.github/config .
+cmake --build build
+
+rm -rf build/
+
```
hihigupt, line 48: this is little ugly, can we make it seperate script?

cmake/build/compile_options.cmake
danewalton: there are bunch of errors in the freertos docs which get flagged. I put them in and grep out the azure failures if I'm checking for doc cleanliness. To check for other clang compiler features, I pull this out and just build since I would always get the freertos failures. Until they (or we) fix the freertos ones this is how I thought about doing it

ports/coreMQTT/azure_iot_core_mqtt.c
danewalton: ahhh yup I should have checked. updated

hihigupt (approved):
:shipit:

PR #124, opened 2021-05-24
Plug and Play Headers

danewalton:
**\~Reference points\~**

John's PR for the embedded header change: https://github.com/Azure/azure-sdk-for-c/pull/1729/files
netxduo JSON reader: https://github.com/azure-rtos/netxduo/blob/master/addons/azure_iot/nx_azure_iot_json_reader.h
netxduo JSON writer: https://github.com/azure-rtos/netxduo/blob/master/addons/azure_iot/nx_azure_iot_json_writer.h


source/include/azure_iot_hub_client.h
```diff
@@ -36,11 +36,15 @@ typedef enum AzureIoTHubMessageQoS
...
-    eAzureIoTHubTwinDesiredPropertyMessage,  /**< The message is a twin desired property message (incoming from the service). */
+    eAzureIoTHubCloudToDeviceMessage = 1,          /**< The message is a cloud message. */
+    eAzureIoTHubDirectMethodMessage,               /**< The message is a direct method message. */
+    eAzureIoTHubCommandMessage,                    /**< The message is a command message. */
+    eAzureIoTHubTwinGetMessage,                    /**< The message is a twin get response (payload contains the twin document). */
+    eAzureIoTHubTwinReportedResponseMessage,       /**< The message is a twin reported property status response. */
+    eAzureIoTHubTwinDesiredPropertyMessage,        /**< The message is a twin desired property message (incoming from the service). */
+    eAzureIoTHubPropertiesGetMessage,              /**< The message is a property get response (payload contains the property document). */
+    eAzureIoTHubPropertiesReportedResponseMessage, /**< The message is a reported property status response. */
+    eAzureIoTHubPropertiesDesiredPropertyMessage,  /**< The message is a desired property message (incoming from the service). */
```
danewalton, line 47: these and the command are the added types

hihigupt: can we just not remove the older names? I am just wondering what happens people used method in application code and we always stamp message in callback as command?

source/include/azure_iot_hub_client.h
```diff
@@ -113,6 +117,24 @@ typedef struct AzureIoTHubClientMethodRequest
     uint16_t usMethodNameLength;   /**< The length of the method name. */
 } AzureIoTHubClientMethodRequest_t;
 
+/**
+ * @brief IoT Hub Command Request struct.
+ */
+typedef struct AzureIoTHubClientCommandRequest
```
hihigupt, line 123: As this is still not public preview... is it ok to just introduce this new concept and remove old concepts? 

@jspaith

jspaith: Yes, we should be 100% all in on PnP concepts.  So any reference to "method" we should have "Command" instead.  Even at public-preview time we can still break stuff, it's only once we hit GA it's a problem.

source/include/azure_iot_json_reader.h
```diff
@@ -0,0 +1,182 @@
...
+/**
+ * @brief Gets the JSON token's number as a 32-bit signed integer.
+ *
+ * @param[in] pxReader A pointer to an #AzureIoTJSONReader_t instance.
+ * @param[out] pilValue A pointer to a variable to receive the value.
+ *
+ * @return An #AzureIoTHubClientResult_t value indicating the result of the operation.
+ * @retval #eAzureIoTHubClientSuccess The number is returned.
+ */
+AzureIoTHubClientResult_t AzureIoTJSONReader_TokenInt32Get( AzureIoTJSONReader_t * pxReader,
```
hihigupt, line 123: Convention followed in all other place is <FileName>_<Verb><>()

her we are ending it with verb. Do we want that?

source/include/azure_iot_json_reader.h
```diff
@@ -0,0 +1,182 @@
...
+    eAzureIoTJSONTokenSTRING = AZ_JSON_TOKEN_STRING,
+    eAzureIoTJSONTokenNUMBER = AZ_JSON_TOKEN_NUMBER,
+    eAzureIoTJSONTokenTRUE = AZ_JSON_TOKEN_TRUE,
+    eAzureIoTJSONTokenFALSE = AZ_JSON_TOKEN_FALSE,
+    eAzureIoTJSONTokenNULL = AZ_JSON_TOKEN_NULL
+} AzureIoTJSONTokenType_t;
+
+typedef struct AzureIoTJSONReader
+{
+    az_json_reader xCoreReader;
```
hihigupt, line 46: Should we have internal anonymous struct, similar to Writer?

source/include/azure_iot_json_reader.h
```diff
@@ -0,0 +1,182 @@
...
+ *
+ */
+
+#ifndef AZURE_IOT_JSON_READER_H
+#define AZURE_IOT_JSON_READER_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "azure_iot_hub_client.h"
```
hihigupt, line 22: I was expecting reader to not depend on any client.

source/include/azure_iot_json_writer.h
```diff
@@ -0,0 +1,293 @@
...
+ *
+ */
+
+#ifndef AZURE_IOT_JSON_WRITER_H
+#define AZURE_IOT_JSON_WRITER_H
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "azure_iot_hub_client.h"
```
hihigupt, line 22: This seems to be wrong dependency

source/include/azure_iot_hub_client_properties.h
```diff
@@ -0,0 +1,308 @@
...
+ * @pre \p xResponseType must be #eAzureIoTHubTwinGetMessage or #eAzureIoTHubTwinDesiredPropertyMessage.
+ * @pre \p pilVersion must not be `NULL`.
+ *
+ * @return An #AzureIoTHubClientResult_t value indicating the result of the operation.
+ * @retval #eAzureIoTHubClientSuccess If the function returned a valid version.
+ */
+AzureIoTHubClientResult_t AzureIoTHubClientProperties_GetPropertiesVersion( AzureIoTHubClient_t * pxAzureIoTHubClient,
+                                                                            AzureIoTJSONReader_t * pxJSONReader,
+                                                                            AzureIoTHubMessageType_t xResponseType,
+                                                                            int32_t * pilVersion );
```
hihigupt, line 213: should the version be uint32_t?

hihigupt: plVersion if it is int32_t*

source/include/azure_iot_hub_client_properties.h
```diff
@@ -0,0 +1,308 @@
...
+
+#ifndef AZURE_IOT_HUB_CLIENT_PROPERTIES_H
+#define AZURE_IOT_HUB_CLIENT_PROPERTIES_H
+
+#include "azure_iot_hub_client.h"
+#include "azure_iot_json_reader.h"
+#include "azure_iot_json_writer.h"
+
+#include <stdbool.h>
+#include <stdint.h>
```
hihigupt, line 24: std header first headers?

source/include/azure_iot_json_writer.h
danewalton: yup you're right copy paste error

source/include/azure_iot_json_reader.h
danewalton: yea I had a reason thinking we needed it to be public but I never fixed. updating

source/include/azure_iot_json_reader.h
danewalton: good catch

PR #129, opened 2021-06-03
add components

danewalton:


source/include/azure_iot_hub_client.h
```diff
@@ -22,6 +22,9 @@
 #include "azure_iot_mqtt_port.h"
 #include "azure_iot_transport_interface.h"
 
+#include "azure/az_core.h"
```
hihigupt, line 25: can we add explicit header for az_iothub_client.h from embedded in this file.?

source/include/azure_iot_hub_client.h
```diff
@@ -30,7 +33,18 @@
...
 
 /* Forward declaration for Azure IoT Hub Client */
-typedef struct AzureIoTHubClient AzureIoTHubClient_t;
+typedef struct AzureIoTHubClient   AzureIoTHubClient_t;
+
+/**
+ * @brief Type for Azure IoT Plug and Play component. The list of component names can be set
+ * as an option for the Azure IoT Hub Client
+ */
+typedef az_span                    AzureIoTHubClientComponent_t;
```
hihigupt, line 42: >                  [](http://example.com/codeflow?start=16&length=17)

nit: please format this properly ... either add one space or make all of the typedef align

Also how about having all MARCO first and then all typedef?

hihigupt (approved):
:shipit:

source/include/azure_iot_hub_client.h
danewalton: Can do the macro one but otherwise it's the uncrustify at work

PR #163, opened 2021-08-02
Update embedded tag and use new function names 

danewalton:
- Updates embedded tag to recently released `1.2.0-beta.1`.
- Updates the function names for a few APIs which were suggested by C SDK core team.


ewertons (approved):

danewalton:
@hihigupt we can probably update the sample repo tag at will if we want. I would suggest doing it at least after the base64 changes I have coming next.

hihigupt (approved):

PR #166, opened 2021-08-04
Fix reported property topic creation

hihigupt:


danewalton (approved):
What a weird oversight....odd we didn't see this behavior in the samples

hihigupt:
> What a weird oversight....odd we didn't see this behavior in the samples

It means our unit tests are busted.... we currently only validate payload send to MQTT. In addition to this we should also validate topic to which we are publishing things to.

PR #167, opened 2021-08-05
Translate core errors to middleware errors

danewalton:


source/azure_iot_internal.c
```diff
@@ -0,0 +1,56 @@
...
+#include "azure/az_core.h"
+
+AzureIoTResult_t _AzureIoT_TranslateCoreError( az_result xCoreError )
+{
+    AzureIoTResult_t xResult;
+
+    switch( xCoreError )
+    {
+        case AZ_OK:
+            xResult = eAzureIoTSuccess;
```
hihigupt, line 16: Right now all these errors are randomly placed and hard to review which errors we are missing to translate. Is it possible to either group using comments ? something like 


 // === JSON error codes ===


  // === HTTP Adapter error codes ===

source/include/internal/azure_iot_internal.h
```diff
@@ -0,0 +1,37 @@
+/* Copyright (c) Microsoft Corporation.
```
hihigupt, line 1: How about putting this file in source folder, and do include using #include "azure_iot_internal.h" in c files. This way none of the build system has to change, be it IDE or cmake as current directory of source is  used for searching header file. 
This way we are telling customers to only include our inc path, while building our stuff. Other option do require them to include internal folder while building our stuff in IDE.

danewalton: If they include the `/source/include` folder this path still resolves since the include strategy should be 
```c
#include "internal/azure_iot_internal.h
```
The embedded SDK does it the same way. This way, if someone is going to include it, they have to explicitly put the `internal` prefix so they know this is an internal header.

hihigupt: having internal in header file name should suffice the intention, I do not think you need folder name also to show that intention. Adding internal folder has downside on useability ... I think we should get embedded sdk to align to this.

source/include/azure_iot_result.h
```diff
@@ -42,7 +42,11 @@ typedef enum AzureIoTResult
     eAzureIoTErrorItemNotFound,          /**< The item was not found. */
     eAzureIoTErrorTopicNoMatch,          /**< The received message was not for the currently processed feature. */
     eAzureIoTErrorTokenGenerationFailed, /**< There was a failure. */
-    eAzureIoTErrorEndOfProperties        /**< End of properties when iterating with AzureIoTHubClientProperties_GetNextComponentProperty(). */
+    eAzureIoTErrorEndOfProperties,       /**< End of properties when iterating with AzureIoTHubClientProperties_GetNextComponentProperty(). */
```
hihigupt, line 45: add section like line 30 and 26

source/azure_iot_provisioning_client.c
```diff
@@ -19,6 +19,7 @@
 #include "azure/az_iot.h"
 #include "azure/core/az_json.h"
 #include "azure/core/az_version.h"
+#include "azure_iot_private.h"
```
hihigupt, line 22: can we sort this list?

source/azure_iot_private.h
```diff
@@ -0,0 +1,25 @@
+/* Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License. */
+
+#ifndef AZURE_IOT_PRIVATE_H
+#define AZURE_IOT_PRIVATE_H
+
+#include "azure_iot_result.h"
+
+#include "azure/az_core.h"
+#include "azure/az_iot.h"
```
hihigupt, line 10: is this header required?

source/azure_iot_json_reader.c
```diff
@@ -11,6 +11,7 @@
 
 #include "azure_iot.h"
 #include "azure_iot_json_reader.h"
```
hihigupt, line 13: >" [](http://example.com/codeflow?start=33&length=1)

can this be first include of  this file as it tries to implement this header.

source/azure_iot_json_writer.c
```diff
@@ -11,6 +11,7 @@
 
 #include "azure_iot.h"
 #include "azure_iot_json_writer.h"
```
hihigupt, line 13: first include of this file.

source/azure_iot_private.h
```diff
@@ -0,0 +1,25 @@
...
+#include <azure/core/_az_cfg_prefix.h>
+
+/**
+ * @brief Translate embedded errors to middleware errors
+ *
+ * @param[in] xCoreError The `az_result` to translate to a middleware error
+ *
+ * @result The #AzureIoTResult_t translated from \p xCoreError.
+ */
+AzureIoTResult_t prvAzureIoT_TranslateCoreError( az_result xCoreError );
```
hihigupt, line 21: >prvAzureIoT_TranslateCoreError [](http://example.com/codeflow?start=17&length=30)

prv is generally used with static function.


"File scope static (private) functions are prefixed with prv."

[https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html](https://www.freertos.org/FreeRTOS-Coding-Standard-and-Style-Guide.html)


How about not adding additional meaning to this prv? Can we name it standard 
xAzureIoT_TranslateCoreError

and let include file name show that it is private ?

source/azure_iot.c
```diff
@@ -13,6 +13,7 @@
 #include "azure_iot.h"
 
 #include "azure_iot_result.h"
+#include "azure_iot_private.h"
```
hihigupt, line 16: >" [](http://example.com/codeflow?start=29&length=1)

sort

source/azure_iot_private.h
danewalton: I'd rather keep the `prv`. The wording there in parenthesis is "private" which I think this function adheres to the essence of the rule. Technically, one could argue it's breaking convention. But fundamentally it would be file specific if it could. Since FreeRTOS doesn't have guidance on this specific scenario, I think it's up to us to decide which camp it belongs in. I think it's closer the `prvAzureIoT` than it is to `xAzureIot`. Plus, since this is internal, it's us that has to deal with the concrete usage and customers would and should only be reading the code. Seeing the keyword `prv` triggers in their head that they should basically ignore.

source/azure_iot_private.h
danewalton: not in the header. I put it in the `.c` now

source/azure_iot_private.h
danewalton: Talked offline with ewerton to get another opinion and he agrees with you. Making the change now

danewalton: Talked offline with ewerton to get another opinion and he agrees with you. Making the change now

hihigupt (approved):
Great Job ! Thanks for taking care of this.

ewertons (approved):

PR #175, opened 2021-08-16
Fix argument type casting on log statements

ewertons:
Logging functions often depend on format translation done similar to
sprintf. In such case, the void* arguments passed need to be cast to
const char* so the format provided matches the args.


ewertons:
This PR replaces https://github.com/Azure/azure-iot-middleware-freertos/pull/173

danewalton (approved):

hihigupt (approved):

PR #178, opened 2021-08-17
directions how to fix code style

danewalton:


danewalton:
Example what this looks like.

image

.github/workflows/ci_tests_linux.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           ( github.event.pull_request.head.repo.fork == true )
 
       - name: Style Check
-        run: .github/scripts/code_style.sh check
+        run: .github/scripts/code_style.sh check || (echo -e '::error There are files which do not adhere to the coding style. Please see directions to fix this here:" https://github.com/Azure/azure-iot-middleware-freertos#code-style' && exit 1)
```
hihigupt, line 31: char(") does not have closing char(")

hihigupt: can you also break this line... so that line is not more then 100 char

hihigupt: space between ( and character

danewalton: The solo `"` is an escape I added a comment detailing that.

hihigupt (approved):

PR #180, opened 2021-08-19
Add azure_iot_config.h template

hihigupt:


danewalton:
I can understand the logging section to be commented out (have the user make a decision). Should the others be uncommented so that there is a default? Or are we having the user make a decision with every one? I'm thinking on the lines of the mbedtls one

hihigupt:
> I can understand the logging section to be commented out (have the user make a decision). Should the others be uncommented so that there is a default? Or are we having the user make a decision with every one? I'm thinking on the lines of the mbedtls one

Default values are already setup using https://github.com/Azure/azure-iot-middleware-freertos/blob/main/source/include/azure_iot_config_defaults.h. Here I am just trying to convey that these values can also be configured.

danewalton (approved):

PR #214, opened 2021-09-29
Update doxygen gate install

danewalton:
Update to version 1.9.0 since 1.8.17 introduced a regression in the `TYPEDEF_HIDES_STRUCT` option. 1.8.17 is the apt-get version for Ubuntu 20.04.
https://github.com/doxygen/doxygen/issues/7508

### DIFF
--- a/.github/scripts/doxygen.sh
+++ b/.github/scripts/doxygen.sh
@@ -8,7 +8,7 @@ set -o nounset # Exit if variable not set.
 set -o pipefail # Exit if pipe failed.
 
 sudo apt update
-sudo apt install -y git cmake python
+sudo apt install -y git cmake python flex bison
 
 pushd ~
 git clone -b Release_1_8_13 https://github.com/doxygen/doxygen.git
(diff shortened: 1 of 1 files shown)


.github/scripts/doxygen.sh
```diff
@@ -8,6 +8,18 @@ set -o nounset # Exit if variable not set.
 set -o pipefail # Exit if pipe failed.
 
 sudo apt update
-sudo apt install -y doxygen
+sudo apt install -y git cmake python flex bison
+
+pushd ~
+git clone -b Release_1_9_0 https://github.com/doxygen/doxygen.git
```
hihigupt, line 14: do we know which feature we are using that is causing us to pin to this version ? If we can remove then we can avoid this building of things.

danewalton: Yea check the link above in the PR. 1.8.17 introduced a regression in that feature (which is crucial) and we therefore are jumping to the next major release. Unfortunately that regression is the same version in apt-get for Ubuntu 20.04. I confirmed at least locally that it works

hihigupt (approved):

PR #224, opened 2022-01-11
update readme

### LINKED ISSUE #221: Initial readme feedback
* In [library architecture](https://github.com/Azure/azure-iot-middleware-freertos#library-architecture), consider calling out that the user can change the TLS|TCP and that the diagram with the mbedTLS|/LWIP.  It's not clear to me right off this is for illustrative purposes and isn't that these are forced.  We don't want to go too much into this in the arch section - if we even take this feedback - since porting guide follows.  
* RE swapping out coreMQTT described [here](https://github.com/Azure/azure-iot-middleware-freertos#mqtt) - we're over-promising on this, at least for now.  We should consider removing this OR call out that the interface here is subject to change until we formalize everything.

danewalton:
closes #221 

adds link to the generated docs on the top level README


README.md
```diff
@@ -49,6 +50,10 @@ For a video deep dive into this repo and the samples, please see the Microsoft D
 
 [<img src="./docs/resources/deep-dive-video.jpg" width="50%">](https://youtu.be/PNykfuJ3VDs)
 
+## Docs
+
+We have published doxygen-generated documentation for all of our APIs. [Please see the link here to review how to use the APIs.](https://azure.github.io/azure-iot-middleware-freertos/)
```
jspaith, line 55: So I really don't know answer here and maybe good topic for embedded side since it has same issue, but... if i weren't familiar with doxygen I'd have no idea to find the reference material by clicking `files`.

Not sure if we should link directly to files.html.  ??

jspaith (approved):

README.md
danewalton: Yea its kind of a trade off. We could link to the files but then what if they were looking for the data structures section etc. I don't mind having it this way. Luckily the top navigation bar is a relative common practice. Maybe if people have issues can revisit.

PR #225, opened 2022-01-12
Fix E2E build

danewalton:
Two things:

1. The newer versions of node drop privileges from child processes. The `chown` makes sure the executables are run as sudo for the network interface acquisition.
2. The E2E tests were hardcoded with a global entry date of `1639093301` which is Thu Dec 09 2021 23:41:41. That has passed and was causing invalid SAS keys to be generated. This change uses the real `time()` to get the unix time.

### DIFF
--- a/tests/e2e/device/e2e_device_commands.c
+++ b/tests/e2e/device/e2e_device_commands.c
@@ -63,7 +63,6 @@ typedef struct E2E_TEST_COMMAND_STRUCT
 typedef E2E_TEST_COMMAND * E2E_TEST_COMMAND_HANDLE;
 /*-----------------------------------------------------------*/
 
-static const uint64_t ulGlobalEntryTime = 1639093301;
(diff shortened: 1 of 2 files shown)


tests/e2e/device/e2e_device_commands.c
```diff
@@ -63,7 +63,6 @@ typedef struct E2E_TEST_COMMAND_STRUCT
 typedef E2E_TEST_COMMAND * E2E_TEST_COMMAND_HANDLE;
 /*-----------------------------------------------------------*/
 
-static const uint64_t ulGlobalEntryTime = 1639093301;
```
danewalton, line 66: @hihigupt do you know why this was originally done this way?

danewalton: Talked offline. This was done for simplicity and because RTC is platform specific. I will be making a change to the other samples and implementations in another PR.

ewertons (approved):

PR #231, opened 2022-05-05
add memcheck

danewalton:
Adds valgrind to the tests.

### DIFF
--- a/.github/scripts/ci_tests.sh
+++ b/.github/scripts/ci_tests.sh
@@ -16,7 +16,7 @@ echo -e "Using FreeRTOS in libraries/FreeRTOS (`git name-rev --name-only HEAD`)"
 TEST_FREERTOS_SRC=`pwd`/libraries/FreeRTOS
 
 echo -e "::group::Building unit tests"
-cmake -Bbuild -DFREERTOS_DIRECTORY=$TEST_FREERTOS_SRC ./tests/ut
+cmake -Bbuild -DFREERTOS_DIRECTORY=$TEST_FREERTOS_SRC -DMEMORYCHECK_COMMAND_OPTIONS="--error-exitcode=1 --leak-check=full" ./tests/ut
 cmake --build build -- --jobs=$TEST_CORES
 pushd build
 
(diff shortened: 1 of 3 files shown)


.github/scripts/ci_tests.sh
```diff
@@ -15,14 +15,17 @@ TEST_RUN_LINE_COVERAGE_THRESHOLD=${4:-80}
...
+echo -e "::group::Building unit tests with Debug"
+cmake -Bbuild -DFREERTOS_DIRECTORY=$TEST_FREERTOS_SRC -DMEMORYCHECK_COMMAND_OPTIONS="--error-exitcode=1 --leak-check=full" -DCMAKE_BUILD_TYPE=Debug ./tests/ut
 cmake --build build -- --jobs=$TEST_CORES
 pushd build
 
 echo -e "::group::Running unit tests"
 ctest -j $TEST_JOB_COUNT -C "debug" --output-on-failure --schedule-random -T test
 
+echo -e "::group::Running unit tests with memcheck"
+ctest -j $TEST_JOB_COUNT -C "debug" --output-on-failure --schedule-random -T memcheck
```
jspaith, line 27: 1) In the original IoT C SDK where we have Valgrind bolted as an option into Cmake, the generated `CTestTestfile.cmake` has Valgrind, Helgrind, and Drd.  Does this all map out to what you've added here?

2) Is there a reason some tests have the `--schedule-random` but others don't?  I'd added that to C SDK way back in day because we thought that the UploadToBlob were bumping into quotas during gate runs where too many would fire at same time (that theory was incorrect fwiw) so want to make sure this is concious on this case.

danewalton: 1. Helgrind and Drd no. But valgrind is there.
1. I think these all have `--schedule-random`? I think we did it just to be sure there is never a chance of inner dependence on sequence of tests (however low likelihood there is of that).

jspaith: Yeah on 2, that was my bad.  I misread one of other lines.  Agree it should be there & be consistent.  Sorry on red herring.

On 1, ;tl-dr => Agree we don't need this for now.  I've just reviewed Helgrind and Drd documentation and given we're single threaded and don't come anywhere near locks I don't see anything they're checking (esp. as fixes usually rely on adding or fixing locks).  On off shot we ever have components of this that can run multi threaded, we'd reconsider this.

jspaith (approved):

Issue #242, opened 2022-08-02
AzureIoTResult return code for AzureIoTHubClient_Connect on refused connection


MarcGeh:
Hello everyone,
I have a feature request. With the current implementation of [AzureIoTHubClient_Connect](https://github.com/Azure/azure-iot-middleware-freertos/blob/ccfda9716f39a70c6772b6fa22f8bf7fecacee90/source/azure_iot_hub_client.c#L736) it is not possible to determine why the connection failed. [xResult](https://github.com/Azure/azure-iot-middleware-freertos/blob/ccfda9716f39a70c6772b6fa22f8bf7fecacee90/source/azure_iot_hub_client.c#L798) is always eAzureIoTErrorFailed regardless of the value of xMQTTResult. However as a user of the API it is necessary for me to tell the difference between a refused connection (MQTTServerRefused) or if the connection failed because of other reasons. In some cases it might be necessary to redo the provisioning if the server refuses the connection. For that purpose I propose the addition of a new enum value for AzureIoTResult that is returned by AzureIoTHubClient_Connect in case the connection got refused.

MarcGeh:
It might be that MQTTServerRefused is the only error that can happen at that point anyways, in that case I could do a re-provisioning simply when I get the eAzureIoTErrorFailed. However that is not really clear to me.

ewertons:
Hi @MarcGeh ,
sorry for the delay.
Unfortunately we will not be able to take this change in.
Please see the reasoning in the linked PR #243

PR #243, opened 2022-08-02
New AzureIoTResult eAzureIoTErrorServerRefused

MarcGeh:
AzureIoTHubClient_Connect returns the new return code eAzureIoTErrorServerRefused if the xMQTTResult is eAzureIoTMQTTServerRefused as suggested by my [feature request](https://github.com/Azure/azure-iot-middleware-freertos/issues/242).


ewertons:
Hi @MarcGeh sorry for the long delay getting back to you.
This unfortunately would be breaking change, and we would not be able to take it within the current major version of the Azure IoT Middleware for FreeRTOS. Because of that we will go ahead and close the PR. Once again, sorry for the delay.

We are designing a new feature to give more specific error reasons, but the work does not get an ETA for completion yet (adding @vaavva for visibility).

PR #246, opened 2022-08-05
Delete download action (deprecated)

danewalton:
Based on the docs, `download` is no longer an action.

https://docs.microsoft.com/en-us/azure/iot-hub-device-update/device-update-plug-and-play#action

### DIFF
--- a/source/include/azure_iot_adu_client.h
+++ b/source/include/azure_iot_adu_client.h
@@ -80,7 +80,6 @@ typedef struct AzureIoTADUClientDeviceProperties
  */
 typedef enum AzureIoTADUAction
 {
-    eAzureIoTADUActionDownload = 0,
     eAzureIoTADUActionApplyDownload = 3,
     eAzureIoTADUActionCancel = 255
 } AzureIoTADUAction_t;


ewertons (approved):

source/include/azure_iot_adu_client.h
```diff
@@ -80,7 +80,6 @@ typedef struct AzureIoTADUClientDeviceProperties
  */
 typedef enum AzureIoTADUAction
 {
-    eAzureIoTADUActionDownload = 0,
```
CIPop, line 83: Is the numerical value important? If not, since this is beta and we don't provide ABI compat we should shift all other values up.

CIPop: We should also have an "unknown" action - i.e. an action that cannot be parsed.
This value could be assigned when parsing fails. Ideally, the state machine can still operate in those conditions by either ignoring or reporting the new action.

danewalton: I had thought about that. Just to align with what is in the docs, I will keep it as is for now and after we release, we can device to add an "unknown" one.

CIPop: Re the linkage - we should then keep the 0 value until it is removed from the spec (but fail accordingly).  It should have a deprecation API doc (similar to how we have in Embedded C SDK). We should also make sure we have all values in the documentation mirrored in our SDK.

Re the unknown: since the values are restricted by the protocol, it's likely better to have only those values. `unknown` will need to be signaled in a different way (e.g. via returned result during parsing).

danewalton: This isn't something we will do right now. If we put those in, it would be after public preview. For now, the sample will treat those deprecated values as "unknown actions"

CIPop (approved):
LGTM for preview.

PR #256, opened 2022-09-29
use latest ubuntu

danewalton:
Moving to `ubunut-latest` aka Ubuntu 20.04 seems to cause a break with nested `npm run` commands (so `npm run alltest` then calls `npm run iothub && npm run prov` etc). This is an unblocker by calling each individually instead of via a catch all `alltest`.


danewalton:
hold on reviewing until this is in PR (not draft)

CIPop (approved):

tests/e2e/service/common/e2e_test_cleanup.ts
```diff
@@ -36,12 +36,14 @@ async function iothubRegistryCleanup(hubConnectionString:string, registryPrefix:
         }
     }
 
-    expiredDevicesDescription.forEach((des) => {
-      console.log("Deleting device: " + des.deviceId)
-    })
-
-    await registry.removeDevices(expiredDevicesDescription, true);
-
+    if(expiredDevicesDescription.length > 0)
```
ericwol-msft, line 39:
```suggestion
    if (expiredDevicesDescription.length > 0)

```

ericwol-msft (approved):

PR #261, opened 2022-10-06
Move JWS validation from samples to source

danewalton:


tests/ut/mbedtls/mbedtls_freertos_port.c
```diff
@@ -0,0 +1,239 @@
+/* Copyright (c) Microsoft Corporation.
+ * Licensed under the MIT License. */
+
+/**
+ * @file mbedtls_freertos_port.c
```
CIPop, line 5: Nit: we should add that this is a test-only implementation.

tests/ut/azure_iot_jws_mbedtls_ut.c
```diff
@@ -0,0 +1,181 @@
...
+
+static int setup( void ** state )
+{
+    memset( ucScratchBuffer, 0, sizeof( ucScratchBuffer ) );
+    return 0;
+}
+
+static void testAzureIoTJWS_ManifestAuthenticate_Success( void ** ppvState )
+{
+    assert_int_equal( AzureIoTJWS_ManifestAuthenticate( ucValidManifest, strlen( ucValidManifest ),
```
CIPop, line 146: I think the expected value should be the first parameter. 
Some UT frameworks will differentiate and the error message when the test fails may be confusing. Even if cmocka doesn't, please take it as a nit (for code readability).

CIPop (approved):
Left a couple of nitpicks.
Thanks @danewalton !

tests/ut/azure_iot_jws_mbedtls_ut.c
danewalton: I hear you. Unfortunately most of the tests (if not all) are this way:

image

PR #270, opened 2022-11-03
Remove root keys from API

danewalton:
- Removes root keys from API. If we want to put them in later we can.


source/interface/azure_iot_jws.h
```diff
@@ -15,6 +15,19 @@
 
 #include "azure_iot_result.h"
 
+/**
+ * @brief Holds the values of the root key used to verify the JWS signature.
+ */
+typedef struct AzureIoTJWS_RootKey
+{
+    uint8_t * pucRootKeyId;
```
CIPop, line 23: What is the root key ID used for?

CIPop (approved):

source/interface/azure_iot_jws.h
danewalton: https://github.com/Azure/azure-iot-middleware-freertos/blob/bf6af17692525dc58e5cc01f8ac1b3d5c11aa00f/ports/mbedTLS/azure_iot_jws_mbedtls.c#L624

The root key has an id within the JWS. We check to make sure the JWS was signed with the root key which we pass to the `AzureIoTJWS_ManifestAuthenticate()` api

CIPop: Shouldn't that check use `pucRootKeyId` from this struct instead?

ports/mbedTLS/azure_iot_jws_mbedtls.c
```diff
@@ -609,7 +609,7 @@ static AzureIoTResult_t prvBase64DecodeJWSHeaderAndPayload( prvJWSValidationCont
     return eAzureIoTSuccess;
 }
 
-static AzureIoTResult_t prvValidateRootKey( prvJWSValidationContext_t * pxManifestContext )
+static AzureIoTResult_t prvValidateRootKey( prvJWSValidationContext_t * pxManifestContext, AzureIoTJWS_RootKey_t * xADURootKey )
```
CIPop, line 612: This function should take an array of RootKeys: we need the `xADURootKeys` and `xADURootKeysCount`.

https://learn.microsoft.com/en-us/azure/iot-hub-device-update/device-update-security#root-keys

> Every Device Update device must contain a set of root keys. These keys are the root of trust for all of Device Update's signatures. Any signature must be chained up through one of these root keys to be considered legitimate.


We should also point to this documentation to make sure application developers are aware that the keys _will change_ over time and key management falls onto the application not the SDK.

> The set of root keys will change over time as it is proper to periodically rotate signing keys for security purposes. As a result, the Device Update agent software will need to be updated with the latest set of root keys at intervals specified by the Device Update team.

source/interface/azure_iot_jws.h
danewalton: Yes sorry this was pasted from the feature branch. In this PR it's changed.

tests/e2e/adu/device/e2e_device_commands.c
```diff
@@ -150,16 +151,60 @@ static uint8_t ucAzureIoTADURootKeyN[ 385 ]
     0x7e, 0x40, 0x48, 0xa9, 0x12, 0x01, 0xa8, 0xd9, 0x7e, 0xf3, 0xa5, 0x1b, 0xf1, 0xfb, 0x90,
     0x77, 0x3e, 0x40, 0x87, 0x18, 0xc9, 0xab, 0xd9, 0xf7, 0x79
     };
-static uint8_t ucAzureIoTADURootKeyE[ 3 ] = { 0x01, 0x00, 0x01 };
+static uint8_t ucAzureIoTADURootKeyE200702[ 3 ] = { 0x01, 0x00, 0x01 };
```
ericwol-msft, line 154: so why specify size? vs below? Compat issue??

static uint8_t ucAzureIoTADURootKeyE200702[] = { 0x01, 0x00, 0x01 };

tests/e2e/adu/device/e2e_device_commands.c
```diff
@@ -150,16 +151,60 @@ static uint8_t ucAzureIoTADURootKeyN[ 385 ]
...
+    0x0b, 0x35, 0xdc, 0x73, 0xc8, 0x69, 0xd0, 0x5e, 0xe8, 0x25, 0x43, 0x9e, 0xf6, 0xd8, 0xab,
+    0x05, 0xaf, 0x51, 0x29, 0x23, 0x55, 0x40, 0x58, 0x10, 0xea, 0xb8, 0xe2, 0xcd, 0x5d, 0x79,
+    0xcc, 0xec, 0xdf, 0xb4, 0x5b, 0x98, 0xc7, 0xfa, 0xe3, 0xd2, 0x6c, 0x26, 0xce, 0x2e, 0x2c,
+    0x56, 0xe0, 0xcf, 0x8d, 0xee, 0xfd, 0x93, 0x12, 0x2f, 0x00, 0x49, 0x8d, 0x1c, 0x82, 0x38,
+    0x56, 0xa6, 0x5d, 0x79, 0x44, 0x4a, 0x1a, 0xf3, 0xdc, 0x16, 0x10, 0xb3, 0xc1, 0x2d, 0x27,
+    0x11, 0xfe, 0x1b, 0x98, 0x05, 0xe4, 0xa3, 0x60, 0x31, 0x99
+};
+static uint8_t ucAzureIoTADURootKeyE200703[ 3 ] = { 0x01, 0x00, 0x01 };
+
+static AzureIoTJWS_RootKey_t xADURootKey[] =
```
ericwol-msft, line 189:
```suggestion
static AzureIoTJWS_RootKey_t xADURootKeys[] =

```

ericwol-msft (approved):

tests/e2e/adu/device/e2e_device_commands.c
danewalton: I think it was mostly to make sure we have them there correctly (size is as it should).

ports/mbedTLS/azure_iot_jws_mbedtls.c
danewalton: Updated. Let me know if you think we should have it changed from the way it's currently done. But I'm going to merge this and continue pulling up changes from the embedded.

PR #271, opened 2022-11-03
add new flash platform function to get the size of the flash bank

vaavva:


source/interface/azure_iot_flash_platform.h
```diff
@@ -17,6 +17,8 @@
 
 AzureIoTResult_t AzureIoTPlatform_Init( AzureADUImage_t * const pxAduImage );
 
+uint32_t AzureIoTPlatform_GetFlashBankSize();
```
danewalton, line 20: thoughts?
```suggestion
size_t AzureIoTPlatform_GetFlashBankSize();
```

danewalton (approved):

source/interface/azure_iot_flash_platform.h
vaavva: Isn't uint32_t more reliably the same size across platforms? Let me know what the pros/cons would be!
I also went with uint32_t because that's the type of ulSizeInBytes that we compare this value against to see which is bigger when accepting/rejecting the update

danewalton: ahhh good point. Usually `size_t` is used for the size of addressable memory on the platform. hmmm this is actually an interesting discussion if we should possibly change that to `size_t`. Realistically, we probably won't have devices larger than 4 GB but who am I to say.

@ewertons | @CIPop thoughts?

CIPop: Ah the `size_t` discussion :). Within Azure SDK for Embedded C we have concluded that we will only use the stdint definitions. Since this is returning flash bank sizes it must be `uint64_t` to allow more than 4GB. (STM32H745I-DISCO ships with a 4GB eMMC today and it's easy to add a >4GB SD card where the flash could be written first).

`size_t` is defined in `stddef.h` and used by the standard library to measure memory length (i.e. that for malloc for example). Note that some smaller MCUs may have `int` and `size_t` different in size (AFAIK PIC18 has 24bit addresses and it's a 16bit MCU).

I agree with @danewalton : `size_t` could make more sense here if we're OK using it in the middleware. Otherwise we should use `uint64_t`.

CIPop, line 20: Nit: naming - should it include "active/passive" or "OTA"? The device will likely have at least 3 banks: boot, bank1, bank2.

CIPop (approved):

source/interface/azure_iot_flash_platform.h
danewalton: We do have precedent for using `size_t` in the middleware.

(private usage)
https://github.com/Azure/azure-iot-middleware-freertos/blob/main/source/include/azure_iot_provisioning_client.h#L87

(public)
https://github.com/Azure/azure-iot-middleware-freertos/blob/b35a1a03d3b33d589f5bedace86a1c38549d5c64/source/interface/azure_iot_mqtt.h#L188

vaavva: Hmm, I would lean away from including OTA since none of the other flash platform functions have it, but it would be good to specify that this is the size of a single boot bank. maybe AzureIoTPlatform_GetSingleFlashBankSize or AzureIoTPlatform_GetFlashBootBankSize? I'll think a bit more on coming up with a concise and clear name

vaavva: Would we need to either change the ulSizeInBytes to uint64_t or size_t, or in the scenario where the function returns uint64_t would we keep it as uint32_t?
```
typedef struct AzureIoTADUUpdateManifestFile
{
    uint8_t * pucId;
    uint32_t ulIdLength;
    uint8_t * pucFileName;
    uint32_t ulFileNameLength;
    uint32_t ulSizeInBytes;
    uint32_t ulHashesCount;
    AzureIoTADUUpdateManifestFileHash_t pxHashes[ AZ_IOT_ADU_CLIENT_MAX_FILE_HASH_COUNT ];
} AzureIoTADUUpdateManifestFile_t;

danewalton: Everything would be `size_t` or `uint64_t` depending on which we settle on.

CIPop: Sorry, I am changing my mind... 

After thinking about file-systems, external storage, etc, I think we should follow the same pattern as `off_t` and `off64_t` which are signed. We should also have <0 checks for invalid values.

> blkcnt_t and off_t shall be signed integer types
(from https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html#tag_13_67)

CIPop (approved):

source/interface/azure_iot_flash_platform.h
vaavva: sync'd offline - best practice decided to be int64_t to allow for systems with external storage and easier discoverability if this value overflows

vaavva: Updated the name - let me know what you think! I leaned towards more descriptive since it's a fairly specific number we're looking for

CIPop: LGTM! Thank you!

ericwol-msft (approved):

PR #279, opened 2022-11-09
Use new embedded unescape

danewalton:


source/azure_iot_adu_client.c
```diff
@@ -215,15 +209,15 @@ AzureIoTResult_t AzureIoTADUClient_ParseRequest( AzureIoTADUClient_t * pxAzureIo
     {
         if( az_span_size( xBaseUpdateRequest.update_manifest ) > 0 )
         {
-            xAzResult = az_json_string_unescape( xBaseUpdateRequest.update_manifest, ( char * ) az_span_ptr( xBufferSpan ), az_span_size( xBufferSpan ), &lOutManifestSize );
+            xBaseUpdateRequest.update_manifest = az_json_string_unescape( xBaseUpdateRequest.update_manifest, xBaseUpdateRequest.update_manifest );
 
             if( az_result_failed( xAzResult ) )
```
ahsonkhan, line 214: This check is no longer required.

tests/ut/azure_iot_adu_client_ut.c
```diff
@@ -745,6 +724,9 @@ static void testAzureIoTADUClient_SendAgentState_WithAgentStateAndRequestWithRet
     uint32_t ulRequestId;
     AzureIoTADUDeviceCustomProperties_t xDeviceCustomProperties;
 
+    /* We have to copy the expected payload to a scratch buffer since it's unescaped in place */
```
CIPop, line 727: Why can't this be left in-place?

tests/ut/azure_iot_adu_client_ut.c
```diff
@@ -334,14 +314,18 @@ static void prvParseRequestSuccess( uint8_t * pucRequestPayload,
...
+    prvParseRequestSuccess( ucRequestPayloadCopy, sizeof( ucADURequestPayload ) - 1,
                             ucADURequestManifest, sizeof( ucADURequestManifest ) - 1 );
 }
 
 /* Test that a request payload with delta update fields parses without errors (ignoring the fields) */
 static void testAzureIoTADUClient_ParseRequest_UnusedFields_Success( void ** ppvState )
 {
-    prvParseRequestSuccess( ucADURequestPayloadUnusedFields, sizeof( ucADURequestPayloadUnusedFields ) - 1,
+    /* We have to copy the expected payload to a scratch buffer since it's unescaped in place */
+    memcpy( ucRequestPayloadCopy, ucADURequestPayloadUnusedFields, sizeof( ucADURequestPayloadUnusedFields ) - 1 );
```
CIPop, line 327: Why do we need to copy?

source/azure_iot_adu_client.c
```diff
@@ -215,15 +209,9 @@ AzureIoTResult_t AzureIoTADUClient_ParseRequest( AzureIoTADUClient_t * pxAzureIo
...
-
-            if( az_result_failed( xAzResult ) )
-            {
-                AZLogError( ( "az_json_string_unescape failed: 0x%08x", xAzResult ) );
-                return AzureIoT_TranslateCoreError( xAzResult );
-            }
+            xBaseUpdateRequest.update_manifest = az_json_string_unescape( xBaseUpdateRequest.update_manifest, xBaseUpdateRequest.update_manifest );
 
-            xAzResult = az_json_reader_init( &jr, az_span_slice( xBufferSpan, 0, lOutManifestSize ), NULL );
+            xAzResult = az_json_reader_init( &jr, xBaseUpdateRequest.update_manifest, NULL );
```
CIPop, line 214: Just to confirm:  are we tracking a service feature request to avoid json encoded base64 json documents?

tests/ut/azure_iot_adu_client_ut.c
```diff
@@ -54,7 +54,7 @@ static uint8_t ucSendStateLongPayloadWithRetry[] = "{\"deviceUpdate\":{\"__t\":\
 
 static uint8_t ucSendResponsePayload[] = "{\"deviceUpdate\":{\"__t\":\"c\",\"service\":{\"ac\":200,\"av\":1,\"value\":{}}}}";
 static uint8_t ucHubClientBuffer[ 512 ];
-static uint8_t ucScratchBuffer[ 8000 ];
+static uint8_t ucRequestPayloadCopy[ 8000 ];
```
CIPop, line 57: Why is this still necessary?

CIPop (commented):
Why do we still need to copy things into a different buffer? The original payload / JSON document should immediately be released by the MQTT stack and I suspect can be modified right?

tests/ut/azure_iot_adu_client_ut.c
danewalton: For the tests we need to. We have an expected payload, which we would like to reuse for different parsing scenarios. After the first test which uses the payload and parses it, the payload gets modified since it's unescaped in place. The next time we try and use that payload for a test, it no longer has an escaped JSON string, but instead has the unescaped in it's place, with a big gap until the end of the string. The JSON parsing then gets screwed up since it has unexpected characters.

tests/ut/azure_iot_adu_client_ut.c
danewalton: reason given above. copy into just for testing purposes.

CIPop (approved):
Thanks @danewalton !

source/azure_iot_adu_client.c
```diff
@@ -215,15 +209,9 @@ AzureIoTResult_t AzureIoTADUClient_ParseRequest( AzureIoTADUClient_t * pxAzureIo
...
         if( az_span_size( xBaseUpdateRequest.update_manifest ) > 0 )
         {
-            xAzResult = az_json_string_unescape( xBaseUpdateRequest.update_manifest, ( char * ) az_span_ptr( xBufferSpan ), az_span_size( xBufferSpan ), &lOutManifestSize );
-
-            if( az_result_failed( xAzResult ) )
-            {
-                AZLogError( ( "az_json_string_unescape failed: 0x%08x", xAzResult ) );
-                return AzureIoT_TranslateCoreError( xAzResult );
-            }
+            xBaseUpdateRequest.update_manifest = az_json_string_unescape( xBaseUpdateRequest.update_manifest, xBaseUpdateRequest.update_manifest );
```
ericwol-msft, line 212: add a comment that this is an inplace operation?

ericwol-msft (approved):

PR #282, opened 2022-11-12
Make use of coreHTTP optional in the sdk

ewertons:
This PR adds a cmake option (OFF by default) to build az::iot_middleware::core_http, not requiring FreeRTOS coreHTTP if customer does not want to use the features that depend on it (as of today, ADU)


.github/scripts/clang_build.sh
```diff
@@ -26,7 +26,7 @@ set -e
 rm -rf build/
 
 echo -e "::group::Build using clang"
-cmake -Bbuild -DCMAKE_C_COMPILER=clang -DFREERTOS_DIRECTORY=$TEST_FREERTOS_SRC -DFREERTOS_PORT_DIRECTORY=$TEST_FREERTOS_SRC/FreeRTOS/Source/portable/ThirdParty/GCC/Posix -DCONFIG_DIRECTORY=.github/config .
+cmake -Bbuild -DCMAKE_C_COMPILER=clang -DFREERTOS_DIRECTORY=$TEST_FREERTOS_SRC -DFREERTOS_PORT_DIRECTORY=$TEST_FREERTOS_SRC/FreeRTOS/Source/portable/ThirdParty/GCC/Posix -DCONFIG_DIRECTORY=.github/config -DUSE_COREHTTP=ON .
```
vaavva, line 29: Is this here for testing that everything works with the flag on? If so, should there be a similar test with it off? (sorry, not super familiar with what's going on in this file)

vaavva (approved):

.github/scripts/clang_build.sh
ewertons: Yes, it is to test the new flag. And we are already testing with and without. This same script builds the solution twice, once for debug and another for release flavor. We are building with coreHTTP only for release here, which is enough to test.

PR #289, opened 2022-11-16
update ADU docs

danewalton:


source/include/azure_iot_adu_client.h
```diff
@@ -387,15 +393,15 @@ AzureIoTResult_t AzureIoTADUClient_SendResponse( AzureIoTADUClient_t * pxAzureIo
...
  * an update request.
- * @param xAgentState The current #AzureIoTADUAgentState_t.
- * @param pxUpdateResults The current #AzureIoTADUClientInstallResult_t. This can be `NULL` if there aren't any
+ * @param[in] xAgentState The current #AzureIoTADUAgentState_t.
+ * @param[in] pxUpdateResults The current #AzureIoTADUClientInstallResult_t. This can be `NULL` if there aren't any
  * results from an update.
- * @param pucBuffer The buffer into which the generated payload will be placed.
- * @param ulBufferSize The length of \p pucBuffer.
- * @param pulRequestId An optional request id to be used for the publish. This can be `NULL`.
+ * @param[in] pucBuffer The buffer into which the generated payload will be placed.
```
vaavva, line 402: Should this one be out?

danewalton: Yea I guess we can. It isn't to be used by the user after this function returns since the API sends it. But it's more of an out than passing data in you're right.

vaavva: There were a few others that seemed to have a similar situation as well from the wording in the comment

source/interface/azure_iot_jws.h
```diff
@@ -20,28 +20,32 @@
...
-    uint8_t * pucRootKeyId;
-    uint32_t ulRootKeyIdLength;
-    uint8_t * pucRootKeyN;
-    uint32_t ulRootKeyNLength;
-    uint8_t * pucRootKeyExponent;
-    uint32_t ulRootKeyExponentLength;
+    uint8_t * pucRootKeyId;           /**< The pointer to the root key id. */
+    uint32_t ulRootKeyIdLength;       /**< The length of the root key id. */
+    uint8_t * pucRootKeyN;            /**< The pointer to the root key modulus. */
+    uint32_t ulRootKeyNLength;        /**< The length of the root key modulus */
```
vaavva, line 26:
```suggestion
    uint32_t ulRootKeyNLength;        /**< The length of the root key modulus. */
```

source/interface/azure_iot_jws.h
```diff
@@ -20,28 +20,32 @@
...
 
-/* This is the minimum amount of space needed to store values which are held at the same time during the calculation. */
-/* azureiotjwsJWS_PAYLOAD_SIZE, one azureiotjwsSIGNATURE_SIZE, and one azureiotjwsSHA256_SIZE are excluded since */
-/* they will reuse buffer space. */
+#define azureiotjwsRSA3072_SIZE                    384                                              /**< Size of the RSA 3072 key. */
+#define azureiotjwsSHA256_SIZE                     32                                               /**< Size of the SHA256 hash. */
+#define azureiotjwsJWS_HEADER_SIZE                 1400                                             /**< Size of the JWS header. */
+#define azureiotjwsJWS_PAYLOAD_SIZE                60                                               /**< Size of the JWS payload. */
+#define azureiotjwsJWK_HEADER_SIZE                 48                                               /**< Size of the JWK header. */
+#define azureiotjwsJWK_PAYLOAD_SIZE                700                                              /**< Size of the JWK payload */
```
vaavva, line 37:
```suggestion
#define azureiotjwsJWK_PAYLOAD_SIZE                700                                              /**< Size of the JWK payload. */
```

vaavva (approved):

source/include/azure_iot_adu_client.h
danewalton: I went through http and crypto as well to do the same thing. Thanks for the callout!

PR #293, opened 2022-11-18
Add note for buffer size

danewalton:
- I got bit by this recently in https://github.com/Azure-Samples/iot-middleware-freertos-samples/pull/279
  - Adding documentation detailing why a minimum buffer size is necessary.
- Also puts all documentation for above the parameters for consistency.


vaavva (approved):
This is great! I've run into this a few times before when I needed to tweak a buffer size, and it does take a bit to debug, so this will definitely reduce that going forward :)

PR #299, opened 2022-11-30
Pull in GA ADU device props fields

danewalton:


vaavva:
Changes addressed in https://github.com/Azure/azure-iot-middleware-freertos/pull/300

Issue #301, opened 2022-12-06
Request for compatibility with ESP-IDF v5.1


wreyford:
Our project uses ESP-IDF 5.1
Would like to be able to just clone the repository, but has required hundreds of hacks to the code.

Of note:
`#include "freertos/FreeRTOS.h"`
`#include "freertos/task.h"`

instead of:
` #include "FreeRTOS.h" `
` #include "task.h" `

Hundreds of format errors with logging eg.:
`AZLogInfo( ( "%.*s", az_span_size( xMessage ), az_span_ptr( xMessage ) ) );`
reports:
` Description	Resource	Path	Location	Type
field precision specifier '.*' expects argument of type 'int', but argument 6 has type 'int32_t' {aka 'long int'} [-Werror=format=]	azure_iot.c	/LoadAssist_IoT/libs/azure-iot-middleware-freertos/source	line 39	C/C++ Problem `
casting to an expected  `(int)az_span_size` solves that.

I've been slowly but surely fixing all these errors, and hopefully will eventually be able to compile and use it.
The only problem is, I will have to repeatedly do this every time you update your repository, so a s to keep up to date with your latest innovations.

This is just a simple request, but also required to keep at pace with Espressif's latest ESP-IDF innovations.

ewertons:
Hi @wreyford ,
We have a PR in progress to make this library compatible with ESP-IDF v5.0 as well.
Hopefully it will suffice to make it compatible with ESP-IDF v5.1 as well. We will not support past ESP-IDF v5.0 yet because that is the latest stable version published by ESPRESSIF.

ewertons:
Hi @wreyford , this PR (#302) has been merged.
If you could share your experience with us, if it addresses your compilation issues, we would appreciate it.

wreyford:
Many hundreds of changes required. 
I started by moving trimming the not required functionality of all the other STM etc support, resulting in many cmake  changes to support new ../../../ depth.

 Besides that, here are the major impactors...I hope I remember them all.

Some of the changes I was required to make to get it to compile with v5.1:
eg. in sample_azure_iot_pnp.c:
`#include "FreeRTOS.h"` to
`#include "freertos/FreeRTOS.h"`
`#include "freertos/task.h"`
This was required in all the classes, also in the libs. I could not figure out a way to not cause this breaking change using cmake, but sure you will get it right, so we don't have to break the code for the other supported architectures.

LogInfo required major changes for v5.1
[...]

PR #302, opened 2022-12-16
Fixes to make azure iot middleware for freertos to work with ESP-IDF v5.0 and Mbed-TLS v3

ewertons:
- Updated library log tracing arguments to match expected data types
  The gcc-extensa compiler in IDF 5 is much more strict and fails if the arguments to log functions (which use VA ARGS) are incompatible with the format string.

- Update azure_iot_jws_mbedtls.c to work with MBED-TLS v3

  Please see this for the differences between mbed-tls v2 and v3:
  https://github.com/Mbed-TLS/mbedtls/blob/development/docs/3.0-migration-guide.md#remove-the-mode-parameter-from-rsa-functions

  To provide compatibility of the ADU JWS code to both mbed-tls v2 and v3 (and also to better comply with RFC 3447 
  https://www.rfc-editor.org/rfc/rfc3447#section-5.1.2), function mbedtls_rsa_pkcs1_verify is now used instead of 
  mbedtls_rsa_pkcs1_decrypt since the device has only access to the public key.

Tested all ESP32 samples (esp32, azureiotkit, adu) using ESP-IDF versions 4.4.3 and 5.

### DIFF
--- a/ports/mbedTLS/azure_iot_jws_mbedtls.c
+++ b/ports/mbedTLS/azure_iot_jws_mbedtls.c
@@ -213,7 +213,7 @@ static AzureIoTResult_t prvJWS_RS256Verify( uint8_t * pucInput,
     /* The signature is encrypted using the input key. We need to decrypt the */
     /* signature which gives us the SHA256 inside a PKCS7 structure. We then compare */
     /* that to the SHA256 of the input. */
-    mbedtls_rsa_init( &ctx, MBEDTLS_RSA_PKCS_V15, 0 );
+    mbedtls_rsa_init( &ctx );
 
     lMbedTLSResult = mbedtls_rsa_import_raw( &ctx,
                                              pucN, ulNLength,
@@ -224,7 +224,7 @@ static AzureIoTResult_t prvJWS_RS256Verify( uint8_t * pucInput,
 
     if( lMbedTLSResult != 0 )
     {
-        AZLogError( ( "[JWS] mbedtls_rsa_import_raw res: %i", lMbedTLSResult ) );
+        AZLogError( ( "[JWS] mbedtls_rsa_import_raw res: %i", ( int ) lMbedTLSResult ) );
         mbedtls_rsa_free( &ctx );
         return eAzureIoTErrorFailed;
     }
@@ -233,7 +233,7 @@ static AzureIoTResult_t prvJWS_RS256Verify( uint8_t * pucInput,
 
     if( lMbedTLSResult != 0 )
     {
-        AZLogError( ( "[JWS] mbedtls_rsa_complete res: %i", lMbedTLSResult ) );
+        AZLogError( ( "[JWS] mbedtls_rsa_complete res: %i", ( int ) lMbedTLSResult ) );
         mbedtls_rsa_free( &ctx );
         return eAzureIoTErrorFailed;
     }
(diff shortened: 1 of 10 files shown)


ports/mbedTLS/azure_iot_jws_mbedtls.c
```diff
@@ -224,7 +225,7 @@ static AzureIoTResult_t prvJWS_RS256Verify( uint8_t * pucInput,
 
     if( lMbedTLSResult != 0 )
     {
-        AZLogError( ( "[JWS] mbedtls_rsa_import_raw res: %i", lMbedTLSResult ) );
+        AZLogError( ( "[JWS] mbedtls_rsa_import_raw res: %i", ( int ) lMbedTLSResult ) );
```
vaavva, line 228: should all the casts in this PR be int_32's/uint_32's to be more explicit?

vaavva (approved):

ports/mbedTLS/azure_iot_jws_mbedtls.c
ewertons: Since that's logging, I'm not so particular about standardizing those. But I will if we have a majority of votes :)

JoseSanchez7 (approved):
I have essentially the same question as Valerie. Does your choosing unsigned ints in certain context have to do with their being able to hold larger positive numbers without extra memory? Or is it just for the sake of diversity and inclusion?

source/azure_iot.c
```diff
@@ -36,7 +36,7 @@ static void prvAzureIoTLogListener( az_log_classification xClassification,
     /* In case logs are stripped out, suppress unused parameter error. */
     ( void ) xMessage;
 
-    AZLogInfo( ( "%.*s", az_span_size( xMessage ), az_span_ptr( xMessage ) ) );
```
danewalton, line 39: lets use the `int32_t` and `uint32_t` instead of `int` and `unsigned int` just to stick with the code style of this and FreeRTOS (if possible).

ports/mbedTLS/azure_iot_jws_mbedtls.c
```diff
@@ -242,45 +243,39 @@ static AzureIoTResult_t prvJWS_RS256Verify( uint8_t * pucInput,
...
         return xResult;
     }
 
-    /* TODO: remove this once we have a valid PKCS7 parser. */
-    shaMatchResult = memcmp( pucBuffer + azureiotjwsPKCS7_PAYLOAD_OFFSET, pucShaBuffer, azureiotjwsSHA256_SIZE );
+    #if MBEDTLS_VERSION_NUMBER >= 0x03000000
+        lMbedTLSResult = mbedtls_rsa_pkcs1_verify( &ctx, MBEDTLS_MD_SHA256, azureiotjwsSHA256_SIZE, pucBuffer, pucSignature );
+    #else
+        lMbedTLSResult = mbedtls_rsa_pkcs1_verify( &ctx, NULL, NULL, MBEDTLS_RSA_PUBLIC, MBEDTLS_MD_SHA256, azureiotjwsSHA256_SIZE, pucBuffer, pucSignature );
+    #endif
```
danewalton, line 265: I think this works temporarily as the PKCS7 structure which ADU uses is simple. But if they decide to make it more complex (using other signed attributes), I think this could break.

[RFC 3852 Section 5.4](https://www.rfc-editor.org/rfc/rfc3852#section-5.4)

> The message digest calculation process computes a message digest on
   either the content being signed or the content together with the
   signed attributes.

I believe PKCS1 is only the message digest so we're lucky here. 

TLDR: Can we keep the `TODO` so we can hopefully change it to PKCS7 in the future if/when they add one?

danewalton (approved):

ports/mbedTLS/azure_iot_jws_mbedtls.c
ewertons: Ok, I'll make the changes.

source/azure_iot.c
ewertons: Ok, I'll update the code.

ports/mbedTLS/azure_iot_jws_mbedtls.c
ewertons: We need to make sure the ADU team gets a feedback about this design, so they take it into consideration. Paraphrasing Cristian, I am surprised the Crypto Board did not raise this flag when they got their design reviewed.

source/azure_iot.c
ewertons: Done.

ports/mbedTLS/azure_iot_jws_mbedtls.c
ewertons: Done.

ewertons:
Related issue: https://github.com/Azure/azure-iot-middleware-freertos/issues/301

PR #307, opened 2023-02-15
Add an issue form for bugs

ewertons:
This might need some adjustments since I could not find a preview viewer.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,57 @@
+name: Bug Report
+description: File a bug report
+title: ""
+labels: []
+assignees:
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for reporting this issue.
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      description: Please search to see if an issue already exists for the bug you encountered.
+      options:
+      - label: I have searched the existing issues
+        required: true
+  - type: textarea
+    id: version
+    attributes:
+      label: Version
+      description: What version of our software are you running? Please provide a tag or commit hash.
+    validations:
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Description of the issue
+      description: Please explain in details the issue observed.
+      placeholder:
+      value:
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behavior
+    attributes:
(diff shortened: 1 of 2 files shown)


ewertons:
For reference: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
+name: Bug Report
+description: File a bug report
+title: ""
+labels: []
```
vaavva, line 4: Opening the file on github has some syntax checking :)
https://github.com/Azure/azure-iot-middleware-freertos/blob/ewertons/issueform/.github/ISSUE_TEMPLATE/bug_report.yml
image

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
...
+      description: What version of our software are you running? Please provide a tag or commit hash.
+    validations:
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Description of the issue
+      description: Please explain in details the issue observed.
+      placeholder:
+      value:
```
vaavva, line 31:
```suggestion
```
I think these could be removed if we're not putting anything in them (since the version field doesn't have them)

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,57 @@
...
+      value: |
+        Thank you for reporting this issue.
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      description: Please search to see if an issue already exists for the bug you encountered.
+      options:
+      - label: I have searched the existing issues
+        required: true
+  - type: textarea
```
vaavva, line 18:
```suggestion
  - type: input
```
Since this will probably be a single line response, input might make the page prettier!

vaavva (approved):
Awesome, after fixing the syntax errors, viewing the file in your branch shows a preview of the form! It looks great!

.github/ISSUE_TEMPLATE/bug_report.yml
ewertons: Fixed. Thanks!

PR #310, opened 2023-05-01
Add ADU for NXP to main readme

vaavva:


ericwol-msft (approved):

README.md
```diff
@@ -41,7 +41,7 @@ To keep this repo simple without board specific dependencies, we have a separate
 - STM32L475 (Azure IoT Hub, Azure Device Update)
 - STM32L4+
 - STM32H745
-- NXP RT1060
+- NXP RT1060 (Azure IoT Hub, Azure Device Update)
```
CIPop, line 44: Nit : AFAIK we support Azure IoT Device Provisioning as well?

CIPop (approved):
Left one small nitpick - applies to all existing sample documentation.

README.md
vaavva: Sounds reasonable to me - would it be Azure IoT Device Provisioning Service, or just Azure IoT Device Provisioning?

ericwol-msft: ugh...is this a table now?

vaavva: That probably does make sense :( All of the boards have Hub and DPS though, right?

vaavva: Changed this to a table - I also checked off the other ST samples that support ADU - should I though since we don't have as much documentation on them?

